PR #43, opened 2014-04-05
Add optional sqs queue output

vijaykramesh:
adds `sqs_queue_name` and `sqs_endpoint` options; if `sqs_queue_name` is specified file deliveries to S3 will also result in a message to SQS indicating the bucket and path of the file delivered. 

we wanted to have a simple way to know which S3 files have been processed downstream after fluentd drops them off to S3, this avoids having to do any sort of pipeline high-water-marking and allows our consumer to just pull files to process off of the SQS queue (without having every single event be it's own item in the queue, as is possible with https://github.com/ixixi/fluent-plugin-sqs

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -102,6 +102,14 @@ The {fluent-mixin-config-placeholders}[https://github.com/tagomoris/fluent-mixin
 
 [utc] Use UTC instead of local time.
 
+[sqs_queue_name] Optional, if included will send JSON message to SQS indicating file delivery to S3, formatted as:
+
+    {
+      "s3_bucket": "YOUR_S3_BUCKET_NAME",
+      "s3_path": "path/to/object"
+    }
+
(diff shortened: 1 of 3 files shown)


ghicks-rmn:
You can get this behavior outside of fluentd using [S3 Events](http://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html). It can be configured to send messages about new S3 objects to SQS, SNS, or Lambda.

okkez:
We can close this issue because [S3 Events](http://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html) is the best way to notify new S3 object to SQS, SNS or Lambda.

How about S3 Events @vijaykramesh ?

vijaykramesh:
thanks for the feedback, we're looking into using S3 Events for this now (and running into performance issues potentially related to this synchronous after_flush)

PR #47, opened 2014-05-24
After write hook

orfah:
This is related to pr https://github.com/fluent/fluent-plugin-s3/pull/43

We implemented a really simple after write hook. Our idea is that the after write scripts would have to have and maintain their own configs, including all the AWS key/secret key configurations. 

I've included a sample after write hook that sends a message to SQS on write, and a config file that would go with it. These are included simply as examples.


repeatedly:
AWS supports S3 Event feature for this purpose so close this PR.
Thanks for your work.

PR #53, opened 2014-10-27
Supports acl options

Hatayaman:
I add acl options to set permission for the object on S3.
It is useful for cross-account access using IAM roles.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -148,6 +148,15 @@ You can change key name by "message_key" option.
 
 [utc] Use UTC instead of local time.
 
+[acl] Permission for the object on S3. It is useful for cross-account access using IAM roles. Valid values are:
+
+- private (default)
+- public_read
+- public_read_write
+- authenticated_read
+- bucket_owner_read
+- bucket_owner_full_control
+
 == IAM Policy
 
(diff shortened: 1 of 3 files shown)


poison:
+1 would be very handy!

ijin:
Ok, so I've looked into this a bit.

At first I thought you can do this with user and bucket policies, but apparently when using cross-account access, the object owner (account B) needs to set an ACL on the object when the bucket owner (account A) is a different account **and** the bucket owner (account A) or whoever else needs to access them.

ref: http://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example3.html

I tested this with IAM roles between 2 accounts, and it seems to work as intended. :+1:

cejast:
is this going to be merged soon? we're currently interested in this

repeatedly:
Need to rebase commits.

jensraaby:
It looks like Hatayaman might not be active on Github (this is the only thing they have committed) - if someone else makes a separate Pull Request it will probably get done faster @repeatedly

jensraaby:
PR #91 is a rebase onto master with conflicts resolved

repeatedly:
New PR based on this PR is merged.

PR #57, opened 2014-11-20
Compression wo copy

kayrus:
If you have 10G buffer log file there is no need to copy it twice. It just consumes additional space and IO operations.


repeatedly:
No response.

PR #68, opened 2015-02-12
Add config_param for AWS S3 Server Side Encryption

alexanderoot:
Added config_param “use_server_side_encryption” to use server-side
encryption with Amazon S3-managed encryption keys

server-side encryption is not enabled by default.


repeatedly:
:+1:

PR #77, opened 2015-05-14
Allow the use of S3 buckets restricted to only the PutObject operation.

JustinHead:
Additionally stores logs on S3 according to creation time and not send time.
Adds the following configuration params:

check_bucket_on_start
logtime
versioning

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -8,6 +8,7 @@ module Fluent
       super
       require 'aws-sdk-v1'
       require 'zlib'
+      require 'date'
       require 'time'
       require 'tempfile'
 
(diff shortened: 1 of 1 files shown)


repeatedly:
`versioning false` may causes log lost with same path.
Is it expected behaviour?

These features introduce code complexity to the plugin but it seems less merit for almost S3 plugin users.
So we can't merge this PR for now.

Maybe, inheriting S3Output and overwriting `ensure_bucket` / `@path_slicer` / `write` resolve your problem.

JustinHead:
Yes, its expected that you know what your doing with the versioning. I agree it adds complexity to the original plugin that not everyone would use.

Are you saying that you would merge if I created perhaps out_s3_putonly.rb and then put a more streamlined version in there by inheriting the base class?

repeatedly:
@JustinHead could be > merge

vrivellino:
Really like the idea behind this PR. I was put off a bit by the recommendation of the docs to [allow all S3 actions](https://github.com/fluent/fluent-plugin-s3#iam-policy):
```json
{ "Statement": [
 { "Effect":"Allow",
   "Action":"s3:*",
   "Resource":"arn:aws:s3:::my-s3bucket*"
  } ]
}
```

That said, I believe this minimal policy will work:
```json
"PolicyDocument" : {
    "Version":"2012-10-17",
    "Statement": [{
        "Effect": "Allow",
        "Action" : [ "s3:Put*", "s3:List*" ],
        "Resource": [
            "arn:aws:s3:::my-s3bucket",
            "arn:aws:s3:::my-s3bucket/*",
        ]
    }]
}
```

Not including `s3:Get*` should vastly help prevent information leakage.

PR #74, opened 2015-03-25
Fix missing credentials error when using IAM Role

sony-shinichi-hatayama:
Missing credentials error occurs when using IAM Role.

According to AWS support, Currently the number of count to retry retrieving credentials is "0".
So I add retry setting.

Please check it.

The stacktrace is the following:

```
2015-03-20 10:34:29 +0000 [error]: failed to configure/start sub output s3: 
Missing Credentials.

Unable to find AWS credentials.  You can configure your AWS credentials
a few different ways:

* Call AWS.config with :access_key_id and :secret_access_key

* Export AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY to ENV

* On EC2 you can run instances with an IAM instance profile and credentials
  will be auto loaded from the instance metadata service on those
  instances.

* Call AWS.config with :credential_provider.  A credential provider should
  either include AWS::Core::CredentialProviders::Provider or respond to
  the same public methods.

= Ruby on Rails

In a Ruby on Rails application you may also specify your credentials in
the following ways:

* Via a config initializer script using any of the methods mentioned above
  (e.g. RAILS_ROOT/config/initializers/aws-sdk.rb).

* Via a yaml configuration file located at RAILS_ROOT/config/aws.yml.
  This file should be formated like the default RAILS_ROOT/config/database.yml
  file.

2015-03-20 10:34:29 +0000 [error]: /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.63.0/lib/aws/core/credential_providers.rb:140:in `credentials'
/opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.63.0/lib/aws/core/credential_providers.rb:62:in `access_key_id'
/opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.63.0/lib/aws/core/client.rb:549:in `build_request'
/opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.63.0/lib/aws/core/client.rb:491:in `block (3 levels) in client_request'
/opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.63.0/lib/aws/core/response.rb:175:in `call'
/opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.63.0/lib/aws/core/response.rb:175:in `build_request'
[...]

PR #79, opened 2015-06-07
Fix usage of environment variables

ijin:
#74 broke the usage of environment variables as configurable credentials. 

This fixes the above problem by specifying `ENVProvider`, which aligns the behavior back to the `DefaultProvider`.
http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/Core/CredentialProviders/DefaultProvider.html

Additionally, the number of retries has been made configurable with `aws_iam_retries`, defaulting to `5` based on the newer implementation of the aws sdk.
http://docs.aws.amazon.com/sdkforruby/api/Aws/InstanceProfileCredentials.html#retries-instance_method

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+~*
+#*
+*~
+[._]*.s[a-w][a-z]
+.DS_Store
+
+*.gem
+.bundle
+Gemfile.lock
+vendor
+.ruby-version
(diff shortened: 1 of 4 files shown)


repeatedly:
Good catch!

repeatedly:
Released v0.5.8.

ijin:
:+1:

PR #81, opened 2015-06-09
Added STS AssumeRole support

ghicks-rmn:
Added support for AssumeRole via two parameters, `sts_role_arn` and `sts_role_session_name`.

AssumeRole can be used in this case to send logs to a bucket owned by a different account. For example, development or staging accounts can forward their logs to the same account as production, allowing all logs to be consumed into the same search or reporting infrastructure.

If a role ARN is found, the plugin takes the existing credentials (IAM role, access/secret key, etc.) and creates a connection to STS, then instantiates an AssumeRole Credential Provider. This provider is passed to the S3 connection instead of the original credentials.

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -20,6 +20,8 @@ module Fluent
     config_param :aws_key_id, :string, :default => nil
     config_param :aws_sec_key, :string, :default => nil
     config_param :aws_iam_retries, :integer, :default => 5
+    config_param :sts_role_arn, :string, :default => nil
+    config_param :sts_role_session_name, :string, :default => nil
     config_param :s3_bucket, :string
     config_param :s3_region, :string, :default => nil
     config_param :s3_endpoint, :string, :default => nil
@@ -80,6 +82,15 @@ module Fluent
       else
         options[:credential_provider] = AWS::Core::CredentialProviders::EC2Provider.new({:retries => @aws_iam_retries})
       end
+
+      # If a role ARN was provided, use the EC2 credential provider to connect to STS.
+      # Then use an AssumeRole provider when connecting to S3.
+      if @sts_role_arn
+        sts_conn = AWS::STS.new(options)
+        provider = AWS::Core::CredentialProviders::AssumeRoleProvider.new({:sts => sts_conn, :role_arn => @sts_role_arn, :role_session_name => @sts_role_session_name})
+        options[:credential_provider] = provider
+      end
+
       options[:region] = @s3_region if @s3_region
       options[:s3_endpoint] = @s3_endpoint if @s3_endpoint
       options[:proxy_uri] = @proxy_uri if @proxy_uri
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -90,12 +90,14 @@ module Fluent
       # Then use an AssumeRole provider when connecting to S3.
       if @sts_role_arn
         sts_conn = AWS::STS.new(options)
-        provider = AWS::Core::CredentialProviders::AssumeRoleProvider.new({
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_s3.rb
```diff
@@ -80,6 +82,15 @@ def start
       else
         options[:credential_provider] = AWS::Core::CredentialProviders::EC2Provider.new({:retries => @aws_iam_retries})
       end
+
+      # If a role ARN was provided, use the EC2 credential provider to connect to STS.
+      # Then use an AssumeRole provider when connecting to S3.
+      if @sts_role_arn
+        sts_conn = AWS::STS.new(options)
```
ijin: Shouldn't we also use other configurable parameters `proxy_uri`, `use_ssl`, etc here?

lib/fluent/plugin/out_s3.rb
```diff
@@ -80,6 +82,15 @@ def start
...
+        provider = AWS::Core::CredentialProviders::AssumeRoleProvider.new({:sts => sts_conn, :role_arn => @sts_role_arn, :role_session_name => @sts_role_session_name})
+        options[:credential_provider] = provider
```
ijin: If we set this, what happens when we are using access key parameters or environment variables? Which takes precedence when passing it to `AWS::S3.new(options)`?

What happens when the assumed role session expires?

lib/fluent/plugin/out_s3.rb
```diff
@@ -20,6 +20,8 @@ def initialize
     config_param :aws_key_id, :string, :default => nil
     config_param :aws_sec_key, :string, :default => nil
     config_param :aws_iam_retries, :integer, :default => 5
+    config_param :sts_role_arn, :string, :default => nil
+    config_param :sts_role_session_name, :string, :default => nil
```
ijin: `:sts_external_id` might also be beneficial

ijin:
Thanks for the pull request!
Overall, I think it's a good feature.

Some accompanying tests would also be great.

lib/fluent/plugin/out_s3.rb
ghicks-rmn: Moved the lines adding these two parameters to the options hash to BEFORE the STS connection is established; this should result in passing the parameters to both the STS and ultimately the S3 connection.

lib/fluent/plugin/out_s3.rb
ghicks-rmn: To avoid ambiguity, I changed this to explicitly delete the access and secret keys from the options hash after establishing the STS connection.

Environment variables are handled by another credential_provider, which we are simply replacing in the options hash before attempting the connection to S3.

lib/fluent/plugin/out_s3.rb
ghicks-rmn: Added support for external_id, with a  default value of nil.

ghicks-rmn:
@ijin I've made some improvements based on your suggestions. I'll try to find time to write tests as well, but I'm leaving for vacation tomorrow and will be out of contact until next month.

ijin:
@ghicks-rmn Thanks for the fixes! LGTM. I can merge as soon as there are some test cases.

repeatedly:
@ijin How about this? Can we merge or need useful tests?

ijin:
@repeatedly The code looks ok, but I would like to have some accompanying tests before merging.

@ghicks-rmn Are you back from vacation?

repeatedly:
@ghicks-rmn ping?

repeatedly:
I can't wait upated patch so we just moved to aws-sdk v2.

okkez:
This is added by #85 , so we can close this PR.

repeatedly:
Closed. Thanks to guys for working on this.

PR #83, opened 2015-06-15
Add timestamp variable

stanhu:
This is useful if you need to want to partition `time_slice` by hour but need
some way of uniquely identifying a file based on creation time.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -51,13 +51,15 @@ Simply use RubyGems:
 - %{time_slice}
 - %{index}
 - %{file_extension}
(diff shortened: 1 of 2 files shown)


repeatedly:
Hmm... I'm not sure this is useful for almost fluent-plugin-s3 users.
file index is not good for?

stanhu:
No, because if the file gets moved out from the directory, index gets reset back to 0. This led to all sorts of confusion and issues for us because the same filename was being used.

repeatedly:
#86 is merged.

Issue #69, opened 2015-02-17
Support for aws-sdk 2.x


edjames:
Hi

Any plans to add support for the 2.x version of the aws-sdk?

repeatedly:
If someone sends a PR.
I was tired from adapting v2 before: https://github.com/fluent/fluent-plugin-s3/issues/51#issuecomment-59319299

edjames:
Ok, I'll do it when I get some time. Thanks.

gerricchaplin:
Hi guys,
Has anyone had any further time to look into this?
Thanks.

edjames:
Hey, I'm still planning to try this. I've now started a work project using Fluentd and now have the time and motivation to do it. If/when I have a working fork I'll raise a PR.

jimwal:
Hi, Is anyone making any progress on this? Cheers.

edjames:
Hi. It's still on my radar but I'm too busy at work to do this right now.
I cannot commit to doing anything right now unfortunately, so please don't rely on me.

I've also changed my workflow so this is not such a priority for me at this point.
Sorry.

Issue #80, opened 2015-06-08
Plugin needlessly requires ListBucketVersions permission


ghicks-rmn:
```
    def check_apikeys
      @bucket.empty?
    rescue AWS::S3::Errors::NoSuchBucket
      # ignore NoSuchBucket Error because ensure_bucket checks it.
    rescue => e
      raise "can't call S3 API. Please check your aws_key_id / aws_sec_key or s3_region configuration. error = #{e.inspect}"
    end
```

This code attempts to run "bucket.empty?" after initializing an S3 connection, ostensibly to check whether the plugin can successfully access the bucket.

What `bucket.empty?` actually does is perform a `ListBucketVersions` call, and returns a boolean depending on whether there were any records in the response.  As a result, the plugin effectively requires permissions to run `ListBucketVersions`, even though this permission has nothing to do with the basic task of uploading messages to S3.

Moreover, there's nothing actually in the documentation (or in the code) to indicate that this permission is required. To discover it, I had to open the plugin source, add debug statements, trace my way through to find the failing line, then check the Ruby SDK source to find out what the method was actually trying to do.

The plugin should be updated to simply print error messages to the log when it fails to upload a file due to permissions.  Or, at the very least, the docs should explicitly say that the `ListBucketVersions` permission is required for the plugin to work at all.

ijin:
I don't know if I would consider it *needlessly*. That's just the way the aws sdk implements bucket availability, as you've discovered. But thanks for digging through and pointing this out.

Pull requests are always welcome. :grinning:

ijin:
Looks like I answered this some time ago. https://github.com/fluent/fluent-plugin-s3/issues/34
(Code has changed, so not everything might be fully relevant)

ghicks-rmn:
First, a correction/clarification: The `ListBucketVersions` permission is only *required* if `check_apikey_on_start` is true. It looks like the SDK returns true if `bucket.exists?` fails due to AccessDenied, which is ultimately the behavior I want - if it can't tell whether the bucket exists, it should assume that it does and try to write to it. I haven't tested but it looks like `ListBucketVersions`would be required for the plugin to actually determine *whether* the bucket exists before attempting to create it - maybe it would be better to simply attempt to create it without even checking, and note in the log if it fails due to AccessDenied or the bucket already existing, as that would not require any additional permissions.

[...]

Issue #25, opened 2013-03-28
Enable it to use by "riak-cs"


m-masaki:
It is made to want you to be able to set up the following option so that "riak-cs" can also be used. 

use_ssl
proxy_uri

Moreover,

It becomes "aws_key_id or aws_sec_key is invalid. Please check your configuration" by "check_apikeys."

Since it is satisfactory, "aws_sec_key" and "aws_sec_key" have been commented out. 

```
@s3 = AWS::S3.new(options)
@bucket = @s3.buckets[@s3_bucket]

ensure_bucket
#check_apikeys
```

ps
英語が苦手なので、日本語でも。
要はriak-csにも送れる様にオプション2つ（use_sslとproxy_uri）をconfigで設定できるようにしてほしいのと、「check_apikeys」で値は間違ってないのにエラーになるので、今はコメントアウトして使ってます。

PR #85, opened 2015-07-10
Use aws sdk v2

okkez:
I'm trying to use aws-sdk v2.

TODOs

- [x] Fix unit test
- [x] Test in real world

See also #69 .


test/test_out_s3.rb
```diff
@@ -311,73 +303,11 @@ def test_auto_create_bucket_false_with_non_existence_bucket
...
-    s3bucket_col.should_receive(:create).with_any_args.and_return { true }
+    setup_mocks
+    @s3_resource.create_bucket(:bucket => "test_bucket")
 
     config = CONFIG_TIME_SLICE + 'auto_create_bucket true'
     d = create_time_sliced_driver(config)
     assert_nothing_raised { d.run }
   end
-
-  def test_aws_credential_provider_default
```
repeatedly: Removing these tests are intented?

okkez: Yes.

See https://github.com/okkez/fluent-plugin-s3/commit/05bf971f5a976e95e9764a80f00e1d9a097c6908

repeatedly: I see. Thanks!

repeatedly:
@ijin Could you also check this PR?

repeatedly:
Sorry, could you rebase the commits?
After that, I will announce this PR for finding testers.

okkez:
Rebased!

lib/fluent/plugin/out_s3.rb
```diff
@@ -15,13 +15,12 @@ def initialize
...
 
     config_param :path, :string, :default => ""
-    config_param :use_ssl, :bool, :default => true
     config_param :use_server_side_encryption, :string, :default => nil
     config_param :aws_key_id, :string, :default => nil, :secret => true
     config_param :aws_sec_key, :string, :default => nil, :secret => true
     config_param :aws_iam_retries, :integer, :default => 5
     config_param :s3_bucket, :string
-    config_param :s3_region, :string, :default => nil
+    config_param :s3_region, :string, :default => "us-east-1"
```
sonots: Should keep with `nil` Otherwise, `ENV['AWS_REGION']` won't be effective.

lib/fluent/plugin/out_s3.rb
```diff
@@ -76,19 +78,15 @@ def start
...
         options[:secret_access_key] = @aws_sec_key
-      elsif ENV.key? "AWS_ACCESS_KEY_ID"
-        options[:credential_provider] = AWS::Core::CredentialProviders::ENVProvider.new('AWS')
-      else
-        options[:credential_provider] = AWS::Core::CredentialProviders::EC2Provider.new({:retries => @aws_iam_retries})
       end
       options[:region] = @s3_region if @s3_region
       options[:s3_endpoint] = @s3_endpoint if @s3_endpoint
       options[:proxy_uri] = @proxy_uri if @proxy_uri
-      options[:use_ssl] = @use_ssl
```
ijin: Looks like we'll be dropping support for riak-cs.
Here's the original context: https://github.com/fluent/fluent-plugin-s3/issues/25

repeatedly: It means AWS SDK v2 doesn't provide `use_ssl` option?

lib/fluent/plugin/out_s3.rb
```diff
@@ -147,8 +146,8 @@ def ensure_bucket
     end
 
     def check_apikeys
-      @bucket.empty?
```
ijin: This probably will affect some users who have minimalistic IAM policies. Furthermore, using `exists?` does not check for access permissions, so I think it defeats the purpose of `check_apikeys`.

See the discussion here: https://github.com/fluent/fluent-plugin-s3/issues/80

lib/fluent/plugin/out_s3.rb
```diff
@@ -76,19 +78,15 @@ def start
       if @aws_key_id && @aws_sec_key
         options[:access_key_id] = @aws_key_id
         options[:secret_access_key] = @aws_sec_key
-      elsif ENV.key? "AWS_ACCESS_KEY_ID"
-        options[:credential_provider] = AWS::Core::CredentialProviders::ENVProvider.new('AWS')
-      else
-        options[:credential_provider] = AWS::Core::CredentialProviders::EC2Provider.new({:retries => @aws_iam_retries})
```
ijin: I believe the original intent here was to have configurable retries for retrieving credentials, because the ec2 metadata service sometimes does not respond reliably. Although `:credential_provider` is no longer available in sdk v2, users can still construct `:credentials` (such as `Aws::InstanceProfileCredentials`) directly with `:retries`. See http://docs.aws.amazon.com/sdkforruby/api/Aws/InstanceProfileCredentials.html

We'll either have to implement something similar to the default provider chain with configurable retries, or drop support for `@aws_iam_retries` from this plugin's options.

test/test_out_s3.rb
ijin: See https://github.com/fluent/fluent-plugin-s3/pull/85/files#r37147150

lib/fluent/plugin/out_s3.rb
ijin: Apparently so. `:endpoint` seems to be the new option.
https://github.com/aws/aws-sdk-ruby/issues/746

From a quick grep search, `proxy_uri` appears to be gone as well.

ijin:
This might break backwards compatibility for some users. A few are attributable to v2 simply not supporting them, but some are from code changes.

I do like the idea of replacing flex-mock with rr, though.

lib/fluent/plugin/out_s3.rb
okkez: Oh! But `Aws::S3::Bucket#empty?` does not exist.
How about `Aws::S3::Bucket#objects.limit(1)`?

lib/fluent/plugin/out_s3.rb
okkez: When keep `nil`, it breaks backward compatibility.
See 4db98de2.

Which is preferred?

lib/fluent/plugin/out_s3.rb
okkez: @repeatedly Which is preferred?

okkez: Add credentials support.
See f692ff7.

lib/fluent/plugin/out_s3.rb
sonots: Okay, then, how about :default => ENV['AWS_REGION'] || 'us-east-1'

okkez: Implemented e59099f.

lib/fluent/plugin/out_s3.rb
okkez: Use `@bucket.objects.first` to check permissions.  2f1102f

lib/fluent/plugin/out_s3.rb
```diff
@@ -68,27 +87,51 @@ def configure(conf)
...
-        options[:credential_provider] = AWS::Core::CredentialProviders::ENVProvider.new('AWS')
+      when @assume_role_credentials
+        c = @assume_role_credentials
+        credentials_options[:role_arn] = c.role_arn
+        credentials_options[:role_session_name] = c.role_session_name
+        credentials_options[:policy] = c.policy if c.policy
+        credentials_options[:duration_seconds] = c.duration_seconds if c.duration_seconds
+        credentials_options[:external_id] = c.external_id if c.external_id
+        options[:credentials] = Aws::AssumeRoleCredentials.new(credentials_options)
+      when @instance_profile_credentials
```
repeatedly: How about using same credentials if user specifies "aws_iam_retries" for backward compatibility?

okkez: Good idea!
I've changed.

repeatedly:
Thanks for applying review!

@ijin How about recent changes? I want to merge this PR and release v0.6.0-pre1 for testing.

ijin:
LGTM.

@okkez Does the required minimum IAM policy change at all?

okkez:
@ijin I'm not sure about IAM policy.
I have not considered required minimum IAM policy.

PR #86, opened 2015-07-20
added new placeholder "uuid_flush" to values_for_s3_object_key: it is…

paolopiaggio:
… replaced with a new uuid everytime the buffer is flushed

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -107,7 +107,8 @@ module Fluent
           "path" => path,
           "time_slice" => chunk.key,
           "file_extension" => @compressor.ext,
-          "index" => i
+          "index" => i,
+          "uuid_flush" => uuid_random
         }
         s3path = @s3_object_key_format.gsub(%r(%{[^}]+})) { |expr|
           values_for_s3_object_key[expr[2...expr.size-1]]
--- a/README.rdoc
+++ b/README.rdoc
@@ -58,6 +58,7 @@ to decide keys dynamically.
 %{time_slice} is the time-slice in text that are formatted with *time_slice_format*.
 %{index} is the sequential number starts from 0, increments when multiple files are uploaded to S3 in the same time slice.
 %{file_extention} is always "gz" for now.
+%{uuid_flush} a uuid that is replaced everytime the buffer will be flushed
 
 The default format is "%{path}%{time_slice}_%{index}.%{file_extension}".
 
--- a/test/test_out_s3.rb
+++ b/test/test_out_s3.rb
@@ -288,6 +288,54 @@ class S3OutputTest < Test::Unit::TestCase
     d.run
   end
 
+  def test_write_with_custom_s3_object_key_format_containing_uuid_flush_placeholder
+    # Assert content of event logs which are being sent to S3
+    s3obj = flexmock(AWS::S3::S3Object)
+    s3obj.should_receive(:exists?).with_any_args.and_return { false }
+    s3obj.should_receive(:write).with(
+      on { |pathname|
+        data = nil
+        # Event logs are compressed in GZip
+        pathname.open { |f|
+          gz = Zlib::GzipReader.new(f)
+          data = gz.read
+          gz.close
+        }
+        assert_equal %[2011-01-02T13:14:15Z\ttest\t{"a":1}\n] +
+                     %[2011-01-02T13:14:15Z\ttest\t{"a":2}\n],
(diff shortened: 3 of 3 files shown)


paolopiaggio:
I've just saw that @stanhu has already submitted a PR that is very similar to this.
This is useful when you want that every file into the destination S3 bucket is generated with a unique name, even if files are removed or moved away from the S3 bucket.
This don't happen with the file index placeholder: the write function tries to write a file with index=0, it a file with that index exists, it tries with index+1 and so on. 
If the file with index=0 was already existent but removed from the S3 bucket, then a new file with index=0 will be stored into the S3 bucket.

stanhu:
Right! I like the idea of the admin being able to choose whether UUID or timestamp is more appropriate.

lib/fluent/plugin/out_s3.rb
```diff
@@ -107,7 +107,8 @@ def write(chunk)
           "path" => path,
           "time_slice" => chunk.key,
           "file_extension" => @compressor.ext,
-          "index" => i
+          "index" => i,
+          "uuid_flush" => uuid_random
```
repeatedly: I will check calling `uuid_random` at each flush is high cost or not.

repeatedly: From benchmark, calling uuid_random at each flush doesn't impact in fluentd performance.

### result

```
% ruby uuid_test.rb
       user     system      total        real
   0.290000   0.000000   0.290000 (  0.293195)
   0.070000   0.000000   0.070000 (  0.064821)
   0.310000   0.000000   0.310000 (  0.310315)
```

### code

```rb
require 'benchmark'
require 'uuidtools'

num = 10000
s3_object_key_format = "%{path}%{time_slice}_%{index}.%{file_extension}"
s3_object_key_format_with_uuid = "%{path}%{time_slice}_%{index}.%{file_extension}.%{uuid_flush}"

Benchmark.bm do |x|
  x.report {
    num.times {
      values_for_s3_object_key = {
        "path".freeze => "logs/".freeze,
        "time_slice".freeze => "1234".freeze,
        "file_extension".freeze => 'txt'.freeze,
        "index".freeze => 0,
        "uuid_flush".freeze => UUIDTools::UUID.random_create.to_s
      }
      s3_object_key_format.gsub(%r(%{[^}]+})) { |expr|
        values_for_s3_object_key[expr[2...expr.size-1]]
      }
    }
  }
  x.report {
    num.times {
      values_for_s3_object_key = {
        "path".freeze => "logs/".freeze,
        "time_slice".freeze => "1234".freeze,
        "file_extension".freeze => 'txt'.freeze,
        "index".freeze => 0
      }
      s3_object_key_format.gsub(%r(%{[^}]+})) { |expr|
        if expr == "%{uuid_flush}".freeze
          UUIDTools::UUID.random_create.to_s
        else
          values_for_s3_object_key[expr[2...expr.size-1]]
        end
      }
    }
  }
  x.report {
    num.times {
      values_for_s3_object_key = {
        "path".freeze => "logs/".freeze,
        "time_slice".freeze => "1234".freeze,
        "file_extension".freeze => 'txt'.freeze,
        "index".freeze => 0
      }
      s3_object_key_format_with_uuid.gsub(%r(%{[^}]+})) { |expr|
        if expr == "%{uuid_flush}".freeze
          UUIDTools::UUID.random_create.to_s
        else
          values_for_s3_object_key[expr[2...expr.size-1]]
        end
      }
    }
  }    
end
```

paolopiaggio:
Cool, thanks @repeatedly!
Do you think you'll plan to merge this PR soon?

repeatedly:
@paolopiaggio Need a test to check uuid is in the path.

paolopiaggio:
done! :)
I don't know ruby very well, but I think the test is ok

repeatedly:
It seems uuid is more safer than timestamp with multiple aggregators.
If there is no problem, I will merge it today.

repeatedly:
Released v0.5.10

paolopiaggio:
Cool! :)
Thanks

sonots:
Can `uuid_flush` placeholder be used to detect duplication of the same chunk? I mean that, when fluent-plugin-s3 mistakenly retries,

1. will `uuid_flush` be changed?
2. will `uuid_flush` be kept same with the previous sending?

If `uuid_flush` behaves in the 2nd way, it is very useful because we can detect mistaken retries, but it looks it works in the 1st way. Can I change the behavior to the 2nd way?

PR #88, opened 2015-07-30
Convert use_server_side_encryption option to symbol when calling AWS::S3

mew109:
When I set *use_server_side_encryption* option, the following error occurred:
**error_class="AWS::S3::Errors::InvalidArgument" error="The encryption method specified is not supported" plugin_id="object:1368d18"**

According to the documentation, the only valid value seems to be *:aes256*
http://www.rubydoc.info/gems/aws-sdk-v1/AWS/Core/Configuration:s3_server_side_encryption
How about changing "use_server_side_encryption" option to a bool?

PR #90, opened 2015-07-31
Support formatter plugin

cosmo0920:
Make `format` option of `out_s3` accept not only `out_file` but also any other formatter plugins.

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -75,9 +75,8 @@ module Fluent
       end
       @compressor.configure(conf)
 
-      # TODO: use Plugin.new_formatter instead of TextFormatter.create
-      conf['format'] = @format
-      @formatter = TextFormatter.create(conf)
+      @formatter = Plugin.new_formatter(@format)
+      @formatter.configure(conf)
 
       if @localtime
         @path_slicer = Proc.new {|path|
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_s3.rb
```diff
@@ -54,9 +54,8 @@ def configure(conf)
       end
       @compressor.configure(conf)
 
-      # TODO: use Plugin.new_formatter instead of TextFormatter.create
-      conf['format'] = @format
-      @formatter = TextFormatter.create(conf)
+      @formatter = Plugin.new_formatter(@format)
```
repeatedly: `Plugin.new_formatter` is available since v0.10.58 so should update gem dependency.

cosmo0920: Should I update gem dependency in this PR?

repeatedly: Yes.

cosmo0920: Thank you for the suggestion!
I updated gem dependency.

repeatedly:
After merged https://github.com/fluent/fluent-plugin-s3/pull/88, I will merge it.

repeatedly:
Could you rebase base on master?

cosmo0920:
Rebased!

cosmo0920:
Thanks!

PR #97, opened 2015-09-08
Allow path based calling format

skizot722:
Some test environments might have various reasons for not supporting the subdomain calling format (it is the default calling format).  This allows enabling of the path-based calling format.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -87,6 +87,8 @@ The {fluent-mixin-config-placeholders}[https://github.com/tagomoris/fluent-mixin
 
     s3_object_key_format %{path}/events/ts=%{time_slice}/events_%{index}-%{hostname}.%{file_extension}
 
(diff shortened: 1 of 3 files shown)


repeatedly:
Thanks!

repeatedly:
I reverted this PR because this option is invalid with AWS SDK v2.
Could you re-send a patch with AWS SDK v2?

skizot722:
Are you sure? I see it documented in v2: http://docs.aws.amazon.com/sdkforruby/api/Aws/Plugins/GlobalConfiguration.html

repeatedly:
Yes. All tests are failed > https://travis-ci.org/fluent/fluent-plugin-s3/builds/79604315

skizot722:
@repeatedly Change is now compatible with AWS v2 SDK with #103 . Original PR was created when fluent-plugin-s3 was using v1 of the AWS SDK, which is why it passed all tests at that time, but tests later failed when fluent-plugin-s3 was switched over to AWS v2 SDK.

PR #96, opened 2015-09-07
Add hex_random placeholder

sonots:
Added `%{hex_random}` placeholder. This is to follow a way of performance tuning written in, `Example 1: Add a Hex Hash Prefix to Key Name` section of a document https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html. 
```
examplebucket/232a-2013-26-05-15-00-00/cust1234234/photo1.jpg
examplebucket/7b54-2013-26-05-15-00-00/cust3857422/photo2.jpg
```

I assume users to use like

```
s3_object_key_format %{hex_random}-%{time_slice}/%{path}_%{index}.json.gz
```

Please see the above linked document for how performance is improved in details.

PR #98, opened 2015-09-09
Add uuid_chunk placeholder => Change the behavior of uuid_flush placeholder to be unique for each buffer chunk

sonots:
I want a uuid for each chunk, not for each flush. 

The purpose is to make it possible to detect duplications occurred by mistaken retries, which might happen if network or s3 states are terrible. 

`%{uuid_flush}` is not available for this purpose, because its uuid is changed on each flush (even on retry).

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -53,6 +53,7 @@ Simply use RubyGems:
 - %{file_extension}
 - %{uuid_flush}
 - %{hex_random}
+- %{uuid_chunk}
 
 to decide keys dynamically.
 
@@ -62,6 +63,7 @@ to decide keys dynamically.
 %{file_extention} is always "gz" for now.
 %{uuid_flush} a uuid that is replaced everytime the buffer will be flushed
 %{hex_random} a random hex string that is replaced for each buffer chunk, not assured to be unique. This is used to follow a way of peformance tuning, `Add a Hex Hash Prefix to Key Name`, written in [Request Rate and Performance Considerations - Amazon Simple Storage Service](https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html). You can configure the length of string with a `hex_random_length` parameter (Default: 4).
+%{uuid_chunk} a uuid that is replaced for each buffer chunk, that is, not replaced on retry unlike `%{uuid_flush}`.
 
 The default format is "%{path}%{time_slice}_%{index}.%{file_extension}".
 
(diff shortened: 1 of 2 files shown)


repeatedly:
How about TLS to keep previous uuid_random value across retry?

sonots:
@repeatedly well, could you elaborate description more?

repeatedly:
You said `%{uuid_flush} is not available for this purpose, because its uuid is changed on each flush (even on retry).` but we can store it in TLS for sharing retries.
This way resolves the problem of uuid change when retry happens.

sonots:
So, use TLS feature instead of using BufferedOutput (TImeSliced) retry. 

If S3 is unstable (network is fine), I doubt this line returns an error https://github.com/sonots/fluent-plugin-s3/blob/uuid_chunk/lib/fluent/plugin/out_s3.rb#L135. TLS retry might not recover this.

repeatedly:
> TLS retry might not recover this.

Why? If so, your uuid_chunk also doesn't recover this problem.

```rb
Thread['previous_uuid_flush'] = uuid_random
# write object
ensure
  Thread['previous_uuid_flush'] = nil
```

I think this code is similar to your uuid_chunk feature.

sonots:
@repeatedly ?? Do you mean use the previous `uuid_flush` on retry? It changes the behavior of uuid_flush. I though guys writing the patch at https://github.com/fluent/fluent-plugin-s3/pull/86 wants to renew the uuid on each flush (not on each chunk). 

If I can change the behavior of uuid_flush as uuid_chunk, it is fine for me.

sonots:
By the way, I first used `Thread[]`, but removed later because I noticed that `chunk.key` is unique for each chunk. We can share the buffer with all threads.

repeatedly:
> wants to renew the uuid on each flush

He wants unique id for each stored files, not consider retry. So changing behaviour is no problem.

sonots:
@repeatedly okay. Then, I will rewrite the behavior of uuid_flush.

repeatedly:
Using `chunk.key` is OK but I don't want to add new parameter.

sonots:
okay, either way is fine. But, let me make sure one thing. Does retry always occur in the same OutputThread? Will the behavior not be changed in the future for sure?

Using `Thread[]` forces only the same thread be able to access the previous uuid_flush.

sonots:
@repeatedly It seems that it is not assured that retry runs on the same thread, so using `Thread[]` won't work http://twitter.com/sonots/status/641484183638970368

sonots:
rebased, but probably will conflict with https://github.com/fluent/fluent-plugin-s3/pull/96. I will rebase again if conflicts.

sonots:
@repeatedly rebased

PR #99, opened 2015-09-09
Add overwrite option to overwrite even if path is duplicated

sonots:
My supposed use case: 

This should be useful to use together with `%{uuid_chunk}` placeholder.
`%{uuid_chunk}` placeholder assures that the path is unique for a chunk,
hence, a path duplication error occurrence means that a mistaken retry
happened. In such case, we probably can simply remove the old file by
overwriting with the new file to fix the state. 

it is difficult to write tests with mock (we can, but it will tell us nothing)

FYI: PR for `%{uuid_chunk}` is https://github.com/fluent/fluent-plugin-s3/pull/98


sonots:
@repeatedly @ijin ping

repeatedly:
Please rebase.

sonots:
@repeatedly rebased

PR #101, opened 2015-09-10
Revert "Allow path based calling format"

repeatedly:
Reverts fluent/fluent-plugin-s3#97


skizot722:
@repeatedly Change is now compatible with AWS v2 SDK with #103 . Original PR was created when fluent-plugin-s3 was using v1 of the AWS SDK, which is why it passed all tests at that time, but tests later failed when fluent-plugin-s3 was switched over to AWS v2 SDK.

PR #102, opened 2015-09-15
Fix a small typo

namusyaka:
This is the same as [this change](https://github.com/fluent/fluentd-docs/pull/195)


repeatedly:
Good catch!

PR #103, opened 2015-09-27
Allow path based calling format

skizot722:
Change is now compatible with AWS v2 SDK.  Original PR was created when fluent-plugin-s3 was using v1 of the AWS SDK, which is why it passed all tests at that time, but tests later failed when fluent-plugin-s3 was switched over to AWS v2 SDK.


repeatedly:
Thanks!

PR #104, opened 2015-09-29
Use `log` instead of `$log`

okkez:
`log` can accepts `log_level` option for each plugin. This change
drops backward compatibility because we support fluentd >= 0.10.58 or
later.


repeatedly:
In `configure` method, using `$log` seems no problem.
This warning messages are important with `log_level error`.

okkez:
Revert partially.
Use `$log` in `configure` method.

PR #111, opened 2015-10-19
Keep hex random identity on rebooting

sonots:
With https://github.com/fluent/fluent-plugin-s3/pull/96, I added `hex_random` placeholder. 
I intended that the `hex_random` values will be identical for same chunks on retrying, but I noticed that

1. `chunk.key` is not unique when data exceeds `buffer_chunk_limit` on the same time_slice. We should've used `chunk.unique_id`
2. Caching the value of `hex_random` on memory does not provide true identity for the same chunk because it is cleared on rebooting fluentd. We should've refetched from file. 

For 2, I decided to use file buffer's tsuffix which is `b5226cf8feb19f2d5` of `tmp/buf.20110102-13.b5226cf8feb19f2d5.log` for example. With this design, we do not need to create another file just to store `hex_random`. What I was disappointed was `FileBufferChunk` does not provide an API to access `tsuffix` directly, though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 Gemfile.lock
 vendor
 .ruby-version
+
+test/tmp/
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ ughs-managing-access-example3.html) for examples.
 The length of `%{hex_random}` placeholder. Default is 4 as written in
 [Request Rate and Performance Considerations - Amazon Simple Storage
 Service](https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html).
+The maximum length is 16.
 
 **overwrite**
 
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -59,6 +59,7 @@ module Fluent
     attr_reader :bucket
 
     include Fluent::Mixin::ConfigPlaceholders
+    MAX_HEX_RANDOM_LENGTH = 16
 
     def placeholders
       [:percent]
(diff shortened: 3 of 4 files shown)


sonots:
For https://github.com/fluent/fluent-plugin-s3/pull/98, `uuid_flush`, we have to do the same thing, but I had no idea how to create identical uuid for an identical chunk. We may should revert #98 simply.

repeatedly:
Ahhh, this patch conflicts master.
Could you rebase this? After that, I will merge it and release new version.

sonots:
@repeatedly I want to discuss for this problem more. What do you think for the identity of `uuid_flush`? Should I simply revert and give it up?

Also, is there a chance to provide `FileBufferChunk#tsuffix` method by Fluentd itself?

repeatedly:
> Also, is there a chance to provide FileBufferChunk#tsuffix method by Fluentd itself?

We don't have such plan because other plugins don't need it.

> Should I simply revert and give it up?

Why? `uuid_flush` seems to not be conflict with `hex_random`.

sonots:
> Why? uuid_flush seems to not be conflict with hex_random.

I wrote about it here > https://github.com/fluent/fluent-plugin-s3/pull/111#issuecomment-149091165

sonots:
> We don't have such plan because other plugins don't need it.

I think `tsuffix` is useful for retry even in other plugins. We may need to design API well, though (If chunk.unique_id is a string, not binary, we could use it).

sonots:
rebased, anyway

repeatedly:
We can't remove `uuid_flush` immediately because it shipped officially.
If we remove it, we should warn for several minor versions.

repeatedly:
> I think tsuffix is useful for retry even in other plugins.

buf_memory doesn't have this value so it seems not useful in retry design.
unique_id based is better.

sonots:
> We can't remove uuid_flush immediately because it shipped officially.

I mean that I changed the behavior to be identical on each chunk, not each flush on #98, but it did not achieve the goal actually (restrating the process will change the uuid). I will send PR to revert #98

PR #116, opened 2015-10-28
Revert "Merge pull request #98 from sonots/uuid_chunk"

sonots:
As discussed in https://github.com/fluent/fluent-plugin-s3/pull/111,
uuid_flush can not be identical for the same chunk on restarting fluentd (the memory cache is cleared).
Let me revert.


sonots:
If someone has a great idea to make `uuid` being identical for the same chunk even on restarting fluentd, please let me know.

PR #120, opened 2015-11-24
Add ssekms_key_id option to use KMS encryption

k0kubun:
I added `ssekms_key_id` option to support server-side encryption by AWS KMS.
This is useful if you want to encrypt secure objects and log who used KMS key to see an object.

Using `:ssekms_key_id` option for `Aws::S3::Object#put`:
http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#put-instance_method


repeatedly:
Released v0.6.2 with this feature.

k0kubun:
Thank you!

PR #122, opened 2015-12-03
Set region to Aws::STS::Client when use Aws::AssumeRoleCredentials

### LINKED ISSUE #119: MissingRegionError when using assume_role_credentials
When using assume_role_credentials, even though s3_region was defined in the configuration file, I got this error:

2015-11-05 01:50:19 +0000 [error]: fluent/engine.rb:172:rescue in run: unexpected error error_class=Aws::Errors::MissingRegionError error=#<Aws::Errors::MissingRegionError: missing region; use :region option or export region name to ENV['AWS_REGION']>

This was after upgrading to latest gem and it also picked up the latest aws sdk.

After editing:
/opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/fluent-plugin-s3-0.6.1/lib/fluent/plugin/out_s3.rb

And adding the :region option, the error was resolved.

okkez:
This may fix #119

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -235,6 +235,9 @@ module Fluent
         credentials_options[:policy] = c.policy if c.policy
         credentials_options[:duration_seconds] = c.duration_seconds if c.duration_seconds
         credentials_options[:external_id] = c.external_id if c.external_id
+        if @s3_region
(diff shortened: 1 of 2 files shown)


okkez:
This PR needs test in real world.

repeatedly:
Thanks. I will test it.

Issue #132, opened 2016-03-07
Documented feature acl not supported


davidhybner:
There are two issues.
First, acl is documented in the Configuration setting, but it doesn't work.  You aren't adding the config_param acl to put_options
```
        put_options[:ssekms_key_id] = @ssekms_key_id if @ssekms_key_id
        put_options[:acl] = @acl if @acl
        @bucket.object(s3path).put(put_options)
```

Second, the documentation is wrong in README.md for acl.  You are using '_' when it should be '-'
Current:
bucket_owner_read
Should be:
bucket-owner-read
See: http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl

repeatedly:
@okkez Could you check?

okkez:
First, `@acl` is not used in out_s3.rb. We should set `@acl` to `put_options`.
Second, current README.md is wrong. We should replace `_` with `-`.

repeatedly:
Release v0.6.6 including fix. Thanks for the report.

PR #134, opened 2016-03-08
Add missing acl option when put object to s3 bucket

okkez:


okkez:
See #132

lib/fluent/plugin/out_s3.rb
```diff
@@ -216,7 +216,12 @@ def write(chunk)
         tmp.rewind
         log.debug { "out_s3: write chunk: {key:#{chunk.key},tsuffix:#{tsuffix(chunk)}} to s3://#{@s3_bucket}/#{s3path}" }
 
-        put_options = {:body => tmp, :content_type => @compressor.content_type, :storage_class => @storage_class}
+        put_options = {
+          :body => tmp,
+          :content_type => @compressor.content_type,
+          :storage_class => @storage_class,
+          :acl => @acl
```
repeatedly: SDK doesn't provide default setting?
If SDK has default, we should not set `acl` when user doesn't specify `acl` parameter?

okkez: [ACL overview](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl) describes ACL default is `private`.
But [SDK document](http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#put-instance_method) does not describes about default value of ACL.

BTW, in previous version S3 bucket and object is created as `private` ACL.

PR #135, opened 2016-03-09
Add warn_for_delay option

sonots:
Add `warn_for_delay` option to output warning logs if delayed events are arrived. The threshold to determine a delay is specified as the value of `warn_for_delay` option.

Assumed scenario is as:

1. s3 objects are put into each daily directory based on **event time** (time field of logs), i.e, `s3_object_key_format` is determined by `time_format`. We want to do so. 
2. We are loading logs from s3 into another data storage such as redshift, BigQuery **daily** using some bulkload tools such as embulk. 
3. We want to notice if delayed events are arrived so that we can refetch old dated directory into the data storage.


repeatedly:
Could you rebase the patch?

sonots:
rebased

PR #138, opened 2016-03-17
Update .travis.yml

sonots:
* drop 1.9.3
* 2.2.0 => 2.2
* add 2.3.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: ruby
 
 rvm:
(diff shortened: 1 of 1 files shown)


repeatedly:
fluent-plugin-s3 itself should work with fluentd v0.10.
So testing with 1.9.3 is needed.

sonots:
Oh, okay, but 1.9.3 is EOL. We should think until when we will support it.

PR #141, opened 2016-03-30
Add 'signature_version' option

mmasaki:
This pathch adds 'signature_version' option.
The signature version 4 is currently used for S3 API request.
However, some S3 compatible storages accept only signature version 2.
The 'signature_version' option enables users to use such a storage.


repeatedly:
thanks!

PR #146, opened 2016-05-06
added csv extension

khetman:
csv is used by many BI tools and better absorbed than pure txt


repeatedly:
Why your commit changes all lines?

PR #149, opened 2016-06-05
Add lz4 compressor

tomfitzhenry:
lz4 is a L77-derivative that achieves better decompression speed than lzo, while achieving the same same compression speed and ratio. According to http://catchchallenger.first-world.info/wiki/Quick_Benchmark:_Gzip_vs_Bzip2_vs_LZMA_vs_XZ_vs_LZ4_vs_LZO .

I have not tested this compressor yet. Looking for early feedback. I'll say when I've tested it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ archive format on S3. You can use serveral format:
 *   text
 *   lzo (Need lzop command)
 *   lzma2 (Need xz command)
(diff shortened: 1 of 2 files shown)


repeatedly:
Sorry, we don't have a plan to add more compressor to s3 plugin.
If you want to share your compressor, release it as a plugin on rubygems.

PR #171, opened 2016-10-19
in: Use block style for `@parser.parse`

okkez:
Because `@parser.parse` without block style is deprecated.

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -218,8 +218,9 @@ module Fluent
       io = @bucket.object(key).get.body
       content = @extractor.extract(io)
       content.each_line do |line|
-        time, record = @parser.parse(line)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_s3.rb
```diff
@@ -218,8 +218,9 @@ def process(body)
       io = @bucket.object(key).get.body
       content = @extractor.extract(io)
       content.each_line do |line|
-        time, record = @parser.parse(line)
-        router.emit(@tag, time, record)
+        @parser.parse(line) do |time, record|
```
repeatedly: Using MultiEventStream is better for performance?

okkez: Ah! Right :+1:
I will use MultiEventStream here.

repeatedly:
Thanks!

PR #173, opened 2016-10-24
Migrate to Fluentd v0.14 API

okkez:


repeatedly:
Sorry for late reply. I need to merge other PR for this patch.
Could you rebase the patch for master?

okkez:
OK I will rebase.

lib/fluent/plugin/in_s3.rb
```diff
@@ -1,115 +1,107 @@
...
-        end
-      end
-    end
-    unless Fluent::Config::ConfigureProxy.method_defined?(:desc)
-      Fluent::Config::ConfigureProxy.class_eval do
-        def desc(description)
-        end
-      end
-    end
+    DEFAULT_PARSE_TYPE = "none"
```
repeatedly: This should be `out_file`: https://github.com/fluent/fluent-plugin-s3/blob/4ed5d9f86ec0b07e41738e6e00398ad88a356b2a/lib/fluent/plugin/out_s3.rb#L97

repeatedly: Ah, sorry. This is input plugin, not output.
My mistake... ;(

repeatedly:
Thanks!

PR #175, opened 2016-11-14
Conditionally check S3 artifacts

chillaxd:
Condition to skip S3 bucket and object existence during runtime, so that more tight and granular AWS policy can be provided on the instance where fluentd is running.

With existing one, we need minimum S3:ListBucket, S3:GetObjectVersion, and S3:PutObject policy. But if we apply this two configure parameter, fluentd will work with only S3:PutObject policy.

For this to happen the below mentioned config parameters must be set:
check_bucket => false
check_object => false


keshavab:
+1
Such conditional checks help put less restrictive policy when using S3 buckets.

repeatedly:
Please add tests.

README.md
```diff
@@ -215,6 +215,14 @@ You can change key name by "message_key" option.
...
 Create S3 bucket if it does not exists. Default is true.
 
+**check_object**
+
+Check object before creation if it exists or not. Default is true.
+
+When it is false,
+	s3_object_key_format will be %{path}/events/%{date_slice}/%{time_slice}.%{file_extension}
+	where, time_slice will be in hhmmss format, so that each object will be unique.
+
```
repeatedly: lack `check_bucket` parameter

repeatedly: And adding s3 policy example is better for this mode.

chillaxd:
After I added the test cases, build got failed. So I delete those test case. Even it is failing. Can you please give some light on this?

repeatedly:
Maybe v0.14.9 changes affect it. I will check it later.

chillaxd:
Thanks Masahiro

chillaxd:
Hi Masahiro, did you got the time to look into the issue?

repeatedly:
I found the cause and I have already opened an PR on fluentd: https://github.com/fluent/fluentd/pull/1319
I will re-check your patch later.

repeatedly:
The test failure are not related your test.
Please add tests again.

chillaxd:
Hi Masahiro, I have added test cases and introduced the mock function to work. Please check

lib/fluent/plugin/out_s3.rb
```diff
@@ -188,35 +194,48 @@ def write(chunk)
...
+          end
 
-        @values_for_s3_object_chunk[chunk.unique_id] ||= {
-          "hex_random" => hex_random(chunk),
-        }
+          i += 1
+          previous_path = s3path
+        end while @bucket.object(s3path).exists?
+      else
+        @s3_object_key_format = "%{path}/%{date_slice}_%{hms_slice}.%{file_extension}"
```
repeatedly: Don't update instance variable in `write`.
Update should be in `configure` or `start`.

lib/fluent/plugin/out_s3.rb
```diff
@@ -142,10 +146,12 @@ def configure(conf)
         @path_slicer = Proc.new {|path|
           Time.now.strftime(path)
         }
+        @hms_slicer = Time.now.strftime("%H%M%S")
```
repeatedly: This value is never updated in `write`.
Is this expected?
All stored object same `@hms_slicer` value.

lib/fluent/plugin/out_s3.rb
```diff
@@ -142,10 +146,12 @@ def configure(conf)
...
       else
         @path_slicer = Proc.new {|path|
           Time.now.utc.strftime(path)
         }
+        @hms_slicer = Time.now.utc.strftime("%H%M%S")
```
repeatedly: ditto

repeatedly:
After updated tests, please squash `fix` like commits into one.

lib/fluent/plugin/out_s3.rb
chillaxd: That's my fault. To refactor I missed that portion.

lib/fluent/plugin/out_s3.rb
chillaxd: Corrected

test/test_out_s3.rb
```diff
@@ -137,6 +147,14 @@ def test_path_slicing_utc
     slice = path_slicer.call(path)
     assert_equal slice, Time.now.utc.strftime("log/%Y/%m/%d")
   end
+  
+  def test_hms_slicing
+    assert_equal '131415', Time.parse("2011-01-02 13:14:15").strftime("%H%M%S")
+  end
+  
+  def test_hms_slicing_utc
```
chillaxd: Removed

repeatedly:
Could you merge "fix" commits into one?
Afte that, I will merge this.

chillaxd:
Hi Masahiro, The previous fixes is what when I try to fix the test cases. And all those changes are in the current commit itself. Do I still need to merge previous "fix" commits into one?

repeatedly:
Yes because if merge current commits into master, master branch has lots of useless `fix` commits.
This is not good for other users, e.g. hard to check changes.

chillaxd:
Done

repeatedly:
Where is the changes of plugin?
Your last change includes only README update.

Issue #176, opened 2016-11-23
Error on adding s3 output to fluentd in container


parkjaeman:
I am trying to add s3 output on fluentd in kubernetes and met following error.
Please let me know what is problem?

-------------------------------------

2016-11-23 05:50:02 +0000 [info]: reading config file path="/etc/td-agent/td-agent.conf"
2016-11-23 05:50:02 +0000 [info]: starting fluentd-0.12.29
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-mixin-config-placeholders' version '0.4.0'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-mixin-plaintextformatter' version '0.2.6'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-elasticsearch' version '1.5.0'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-kafka' version '0.3.5'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-kafka' version '0.3.1'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-mongo' version '0.7.15'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-rewrite-tag-filter' version '1.5.5'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-s3' version '0.7.2'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-s3' version '0.7.1'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-scribe' version '0.10.14'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-td' version '0.10.29'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-td-monitoring' version '0.2.2'
2016-11-23 05:50:02 +0000 [info]: gem 'fluent-plugin-webhdfs' version '0.4.2'
2016-11-23 05:50:02 +0000 [info]: gem 'fluentd' version '0.12.29'
2016-11-23 05:50:02 +0000 [info]: adding match pattern="fluent.**" type="null"
2016-11-23 05:50:02 +0000 [info]: adding match pattern="*.**" type="s3"
2016-11-23 05:50:02 +0000 [error]: unexpected error error="No such file or directory - addr"
[...]

PR #177, opened 2016-11-24
Cleanup dependency

repeatedly:
This also fixes https://github.com/fluent/fluent-plugin-s3/issues/176

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source "http://rubygems.org"
 
 gem 'json', '= 1.8.3'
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -1,5 +1,7 @@
 source "http://rubygems.org"
 
 gem 'json', '= 1.8.3'
+gem 'uuidtools'
+gem 'fluentd', '= 0.14.8'
```
repeatedly: This is workaround for v0.14.9 regression.

PR #182, opened 2017-01-26
Allow ignore s3 endpoint certificate verification

cmarodrigues:
When we define s3 endpoint with self signed certificate, we get errors of certificate invalid.

This is useful for integration with s3 endpoints of EMC/Dell Centera S3 API.


repeatedly:
Please add it to README.

cmarodrigues:
Done.

PR #185, opened 2017-02-21
Support http-wire-trace for debugging

okkez:


lib/fluent/plugin/in_s3.rb
```diff
@@ -185,19 +190,30 @@ def create_s3_client
       options = setup_credentials
       options[:region] = @s3_region if @s3_region
       options[:proxy_uri] = @proxy_uri if @proxy_uri
+      log.trace do
```
repeatedly: How about `if log.level == Fluent::Log::LEVEL_TRACE`?

tagomoris: `Fluent::Log#on_trace` method is defined for such case.

repeatedly: Ah, I forgot it.

okkez: Ah, I forgot `Fluent::Log#on_trace`. Thanks!

PR #186, opened 2017-03-02
Fix placeholder handling

repeatedly:
Fix the problem that placeholders can't be used in `s3_object_key_format`.

In addition, keep existing behaviour: time slice is day range


tpetr:
💯

repeatedly:
Released v1.0.0.rc3

PR #193, opened 2017-05-10
Add missing hms_slice object key

cosmo0920:
This regression had been introduced in
f10df85afd76f070a059c9e5ca8d97bdd3323cd7.


repeatedly:
Thx!

PR #196, opened 2017-06-16
CGI unescape S3 key

IvanVergiliev:
From http://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html:

> The s3 key provides information about the bucket and object involved in the event. Note that the object keyname value is URL encoded. For example "red flower.jpg" becomes "red+flower.jpg".

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -2,7 +2,7 @@ require 'fluent/plugin/input'
 require 'fluent/log-ext'
 
 require 'aws-sdk-resources'
-require 'cgi'
+require 'cgi/util'
 require 'zlib'
 require 'time'
 require 'tempfile'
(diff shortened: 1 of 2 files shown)


repeatedly:
@okkez Could you check this?

okkez:
OK, I will check this PR later.

lib/fluent/plugin/in_s3.rb
```diff
@@ -2,6 +2,7 @@
 require 'fluent/log-ext'
 
 require 'aws-sdk-resources'
+require 'cgi'
```
okkez: Use `require 'cgi/util'`, please.
Becasue `require 'cgi'` loads all libraries related to CGI, but we don't use all cgi feature.

repeatedly:
Thanks!

PR #197, opened 2017-06-19
Fix failed s3-input caused by url encoded s3 object key

kei-q:
If S3 object key includes URL encoded characters, then this plugin failed to fetch S3 object.

```
2017-06-19 13:57:06 +0900 [warn]: #0 plugin/in_s3.rb:147:rescue in block in run:  error_class=Aws::S3::Errors::NoSuchKey error="The specified key does not exist."
```

s3 object key received from SQS need `URI.decode` before to fetch S3 object.

for example:

original path : `s3://sample-bucket/dt=2017-06-19/sample-file.json`
`s3["object"]["key"]` : `s3://sample-bucket/dt%3D2017-06-19/sample-file.json`
`URI.decode(s3["object"]["key"])` : `s3://sample-bucket/dt=2017-06-19/sample-file.json`

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -219,8 +219,9 @@ module Fluent::Plugin
     def process(body)
       s3 = body["Records"].first["s3"]
       key = s3["object"]["key"]
+      decoded_key = URI.decode(key)
 
-      io = @bucket.object(key).get.body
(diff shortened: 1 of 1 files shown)


kei-q:
I close this PR, because it duplicated : https://github.com/fluent/fluent-plugin-s3/pull/196

PR #202, opened 2017-08-17
Add use_bundled_cert parameter to use ruby-sdk-ruby bundled cert

repeatedly:
If the OS / middleware provided cert has a problem with AWS API, need to use bundled cert instead.


repeatedly:
@mururu Could you check this?

mururu:
@repeatedly Looks good 👍

PR #203, opened 2017-08-23
support adding arbitrary S3 metadata on objects

tpetr:


lib/fluent/plugin/out_s3.rb
```diff
@@ -110,6 +110,8 @@ def initialize
     config_param :signature_version, :string, default: nil # use nil to follow SDK default configuration
     desc "Given a threshold to treat events as delay, output warning logs if delayed events were put into s3"
     config_param :warn_for_delay, :time, default: nil
+    desc "Arbitrary S3 metadata headers to set for the object"
+    config_param :s3_metadata, :hash
```
repeatedly: The default value should be `nil`

tpetr: Updated, thanks

repeatedly:
Thanks!

PR #206, opened 2017-09-13
Fix s3 logging

tpetr:
- Wrapping the log statement in `{ }` was causing it not to actually be logged
- `chunk.key` and `tsuffix()` don't seem to exist

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -268,7 +268,7 @@ module Fluent::Plugin
       begin
         @compressor.compress(chunk, tmp)
         tmp.rewind
-        log.debug "out_s3: write chunk: {key:#{chunk.key},tsuffix:#{tsuffix(chunk)}} to s3://#{@s3_bucket}/#{s3path}"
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_s3.rb
```diff
@@ -268,7 +268,7 @@ def write(chunk)
       begin
         @compressor.compress(chunk, tmp)
         tmp.rewind
-        log.debug { "out_s3: write chunk: {key:#{chunk.key},tsuffix:#{tsuffix(chunk)}} to s3://#{@s3_bucket}/#{s3path}" }
+        log.debug "out_s3: write chunk with metadata #{chunk.metadata} to s3://#{@s3_bucket}/#{s3path}"
```
tpetr: @repeatedly I realized that `chunk.metadata` might only exist for 0.14+ -- is there a better thing to print that will give useful info for all versions?

repeatedly: Latest version, 1.0.0 or later, is only for v0.14, so considering v0.12 is not needed I think.

PR #210, opened 2017-10-16
log chunk ID and support %{index} placeholder in S3 metadata

tpetr:
Related PR for chunk ID logging: https://github.com/fluent/fluentd/pull/1718


repeatedly:
Thanks!

Issue #34, opened 2013-10-15
What are the minimum permissions on S3 policy?


gaia:
I want to attach the most restrictive policy possible to the key used by the plugin. I have this fairly restrictive policy in place, and it seems to work.

Am I missing any actions that I should allow?

Are there any actions I can remove from the list below?

`{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "s3:AbortMultipartUpload",
        "s3:CreateBucket",
        "s3:ListAllMyBuckets",
        "s3:ListBucket",
        "s3:ListBucketMultipartUploads",
        "s3:ListBucketVersions",
        "s3:ListMultipartUploadParts",
        "s3:PutObject"
      ],
      "Sid": "Stmt1XXXXXXXXXXXXX",
      "Resource": [
        "arn:aws:s3:::myfluentdbucket"
      ],
      "Effect": "Allow"
    }
  ]
}`

bryanlarsen:
For the current version you also have to set `check_object false` or give the GET_OBJECT permission.

PR #216, opened 2017-11-08
Update IAM policy example

ryysud:
I found that the existing example is not a minimal IAM policy.
So, I fix it by referring to this issue (https://github.com/fluent/fluent-plugin-s3/issues/34) .


repeatedly:
Thanks!
I forgot to update the README...

Issue #54, opened 2014-10-28
Optional zero padding for index field


nick-peskett-bbc:
I'm running into problems with an installation which is using a daily time_slice_format and uploading a flush_interval of 5 minutes.

The problem is due to the only sorting available with GET Bucket (List Objects) being alphabetical.

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html

I'd very much like to use the marker feature when listing the objects in the bucket, but (for example) if the last file I downloaded was 20140925_11.gz, using this as a marker would mean files 12-109 (uploaded in the meantime) would be missed.

An option like "zero_padding_length 4" would be really useful in this example.

Order returned when listing bucket;

[mod date]          filename
[24/09/14 23:04:56] 20140925_0.gz
[24/09/14 23:09:50] 20140925_1.gz
[24/09/14 23:54:58] 20140925_10.gz
[25/09/14 07:26:30] 20140925_100.gz
[25/09/14 07:31:32] 20140925_101.gz
....
[25/09/14 08:06:39] 20140925_108.gz
[25/09/14 08:11:39] 20140925_109.gz
[24/09/14 23:59:59] 20140925_11.gz
[25/09/14 08:16:41] 20140925_110.gz

PR #219, opened 2018-01-05
Format index with `sprintf`

JamesJJ:
Possible solution for
https://github.com/fluent/fluent-plugin-s3/issues/54


repeatedly:
Thanks!

PR #220, opened 2018-01-12
in_s3: Add bucket and object key to parsed records

gregsymons:
My use case for this is parsing logs from cloudfront and s3 that have all been dumped in a single bucket at different prefixes. Since cloudfront and s3 have different logging formats, I need a way to process the records based on what type they are. Adding the key lets me use that to retag the record using the record_transformer filter and then use that tag further down the pipeline.

I could see a use case for adding other metadata from the object to the record, but I don't need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,11 +593,16 @@ Given a threshold to treat events as delay, output warning logs if delayed event
       aws_sec_key YOUR_AWS_SECRET_KEY
       s3_bucket YOUR_S3_BUCKET_NAME
       s3_region ap-northeast-1
+      add_object_metadata true
 
       <sqs>
         queue_name YOUR_SQS_QUEUE_NAME
       </sqs>
     </source>
+
+**add_object_metadata**
+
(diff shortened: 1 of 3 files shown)


repeatedly:
Sorry for the late reply.
We need to add parameter to control this behaviour because adding new fields may break existing users e.g. schema mismatch.
So adding `add_target_info` like paramter is better.

gregsymons:
That seems reasonable. I'll add that and update the PR.

gregsymons:
Finally got back to this. Added a configuration option and updated the tests to test with it both on and off.

TaLoN1x:
does this fix allows using variables like ${tag} in the bucket name?

gregsymons:
@TaLoN1x No, this fix only augments the parsed record with the bucket and key.

kochp:
Great feature! I just successfully tried it to solve my case. I there any reason why it is not merges to master? Would be helpful in the official version of this plugin.

repeatedly:
This is our mistake. Forgot to merge. Will release new version sooon.

PR #225, opened 2018-02-28
Added document for `use_bundled_cert` parameter

kenfdev:
I've been trying to use fluentd on a Windows environment and kept on getting the error as below.

```
SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (Seahorse::Client::NetworkingError)
```

I did some research and found PR #202 where the `use_bundled_cert` parameter has been added.
It would be helpful if this is documented on the README.md, hence I added it.
Thanks!


repeatedly:
Ah, good catch!
Thanks.

PR #226, opened 2018-03-28
in_s3 - gzip - Fixed extracting multiple streams

rudolf-focuscura:
Fixed extracting of multiple concatenated compressed streams.
https://www.gnu.org/software/gzip/manual/html_node/Advanced-usage.html

We found this issue after spending a couple of days investigating missing AWS ELB & ALB logs (which are apparently in this format).

Ps.: I don't have too much experience with Ruby, but I hope this looks okay.


rudolf-focuscura:
Should I try to cover this with an unit test?

repeatedly:
Yeah, adding test is better.
Creating response body with gzipped content can test it: https://github.com/fluent/fluent-plugin-s3/blob/ed507dffe902eb8ba026daf0f3e80714c30b5d1b/test/test_in_s3.rb#L162

rudolf-focuscura:
I finally managed to find time to finish this, sorry for the long delay.

rudolf-focuscura:
Rebased on `master` and added `Signed-off-by` to commit messages.

PR #227, opened 2018-04-10
Fix memory leak when `s3_object_key_format` contains time-varying placeholder

eagletmt:
Since `Fluent::Plugin::Output#extract_placeholders` uses format string
as cache key [1], expanding time-varying placeholders before
`extract_placeholders` leads to memory leak bug. In fluent-plugin-s3,
`%{path}` and `%{file_extension}` are fixed values specified in
configuration file, but other placeholders like `%{time_slice}` are
time-varying.

[1]: https://github.com/fluent/fluentd/blob/v1.1.3/lib/fluent/plugin/output.rb#L708


repeatedly:
Good catch! Thanks.

PR #238, opened 2018-07-13
Enable time_slice in s3_object_key_format even when check_object is false

south37:
## Why
When `check_object` is false, `%{time_slice}` cannot be used in `s3_object_key_format` and `%{date_slice}` should be used instead of it.
I think this behavior is very confusing.

## What
This PR enables the usage of `%{time_slice}` in `s3_object_key_format` even when `check_object` is false.

### Before
fluent config is below.

```fluentd
<match pattern>
  @type s3

  .
  .
  .
  check_object false
  s3_object_key_format %{path}/%{time_slice}_%{hms_slice}.%{file_extension}
</match>
```

s3 object path will be `s3://.../_112020.gz`.

### After

```fluentd
<match pattern>
  @type s3

  .
  .
  .
  check_object false
  s3_object_key_format %{path}/%{time_slice}_%{hms_slice}.%{file_extension}
</match>
```

s3 object path will be `s3://.../20180713_112020.gz`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,8 +299,8 @@ When it is false,
 Check object before creation if it exists or not. Default is true.
 
 When it is false,
-	s3_object_key_format will be %{path}%{date_slice}_%{time_slice}.%{file_extension}
-	where, time_slice will be in hhmmss format, so that each object will be unique.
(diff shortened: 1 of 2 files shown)


repeatedly:
Looks good. Thanks!

south37:
Thank you for your reviewing! 😄

PR #242, opened 2018-09-05
Improve check_apikeys() perfomance.

tkimura:
If there are too many files under @path, it may take several tens of seconds to execute check_apikeys().

By specifying max_keys, it will be able to run in less than 1 second regardless of the number of files.

reference:

* https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Bucket.html#objects-instance_method


repeatedly:
Ah, good catch 👍 
Thanks!

tkimura:
Thanks to merge !

Is this changes merge to v0.12 branch?
I'm use td-agent 0.12.40 now.

repeatedly:
> Is this changes merge to v0.12 branch?

Hmm... could you send same patch to v0.12 branch?

tkimura:
I'm created new PR.
https://github.com/fluent/fluent-plugin-s3/pull/243

PR #244, opened 2018-09-10
Add endpoint support to in_s3

duck8823:
I tried this plugin with locakstack for testing and checking, but did not work.
`out_s3` plugin already have parameter `s3_endpoint`, but `in_s3` plugin not.  
Thus, I just added some parameters to set the endpoint.

Configurations can be set as follows
```
<source>
  ...
  s3_endpoint http://localstack:4572
  force_path_style true
  <sqs>
    queue_name my-queue
    endpoint http://localstack:4576
  </sqs>
</source>
```

I think that it will be easier to test and check locally, if these parameters become available.


repeatedly:
Thanks!

PR #254, opened 2018-11-14
Update proxy setting for S3 input plugin

jayhding:
According to https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html, the proxy option name for S3 should be `http_proxy` instead of `proxy_uri`. It should fix below error

```
2018-11-14 22:35:21 +0000 [error]: #0 unexpected error error_class=ArgumentError error="invalid configuration option `:proxy_uri'"
```

Also, `http_proxy` is missing in `create_sqs_client` function


repeatedly:
This is our mistake. Thanks!

PR #259, opened 2018-12-05
Fix line breaks in README.md

opklnm102:


README.md
```diff
@@ -282,11 +282,9 @@ to decide keys dynamically.
...
 E.g., "logs/" in the example configuration above.
 * %{time_slice} is the
-time-slice in text that are formatted with **time_slice_format**. %{index} is
-the sequential number starts from 0, increments when multiple files are
-uploaded to S3 in the same time slice.
-* %{file_extention} is always "gz" for
-now.
+time-slice in text that are formatted with **time_slice_format**.
+* %{index} is the sequential number starts from 0, increments when multiple files are uploaded to S3 in the same time slice.
+* %{file_extention} is always "gz" for now.
```
repeatedly: This is outdated because it depends on `Compressor`.
So `%{file_extension} depends on **store_as** parameter` is better.

opklnm102: I guess the %{index} is not correct for line breaks.
It looks like you need a fix.

opklnm102:
Thanks to merge!

PR #265, opened 2019-01-27
in_s3: Catch SQS poller exception and restart [retry] the poller loop

JamesJJ:
The `@poller.poll(...)` block is supposed to run forever unless shutdown. 

This PR adds a retry to catch when poll throws any exception and so keeps the polling thread running.

e.g. Network issue between fluentd and AWS, would cause `.poll` to throw.


repeatedly:
Looks good.

PR #266, opened 2019-02-27
Pass chunk directly instead of chunk.metadata

cosmo0920:
extract_placeholders should be passed `chunk` instance directly instead of `chunk.metadata`.

Signed-off-by: Hiroshi Hatake <cosmo0920.oucc@gmail.com>


repeatedly:
Thanks!

PR #285, opened 2019-08-19
remove a feature request leaking into the README

mburns:
reverts #188

... which added what appears to be a feature request in the form of a sub-heading in the project readme.

The readme doesn't seem like the right place for this (maybe better as an Issue), and is confusing to readers trying to learn about the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,6 @@ alt="Build Status" />](https://travis-ci.org/fluent/fluent-plugin-s3) [<img
 src="https://codeclimate.com/github/fluent/fluent-plugin-s3/badges/gpa.svg"
 />](https://codeclimate.com/github/fluent/fluent-plugin-s3)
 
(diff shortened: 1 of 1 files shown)


mburns:
rfr @ganmacs

ganmacs:
> The readme doesn't seem like the right place for this (maybe better as an Issue)

thank you for your contribution. You're right.  Look good to me. 
BTW, could you follow DCO? https://github.com/fluent/fluent-plugin-s3/pull/285/checks?check_run_id=197460290

@repeatedly 
what do you think of it? I completely agree with this change. I think that these kinds of information should be at issue, not README. if you don't have any concerns, I'll merge this and create a new issue about Ceph s3.

repeatedly:
No problem for me

ganmacs:
thank you!!

PR #286, opened 2019-08-23
Add possibility to configure bucket lifecycle policy rules

coder-hugo:
This pull request adds the possibility to configure [lifecycle rules](https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lifecycle-mgmt.html) for the bucket. Lifecycle rules are useful if you don't want to store the logs forever in the s3 bucket. With such rules the S3 cleans up old logs automatically. At the moment we have to do this manually for all our log bucket although there where created automatically by this plugin.


repeatedly:
Ah, I forgot to merge this PR.
Will release new version soon.

Issue #290, opened 2019-10-03
Add the actual Object metadata when add_object_metadata is enabled


hlx98007:
AWS s3 object has customizable metadata, the aws CLI version of getting the custom metadata is located here: https://docs.aws.amazon.com/cli/latest/reference/s3api/head-object.html

While I think the bucket & key is useful but it would be more powerful to add the custom metadata into this configuration.

sppwf:
Hi Guys,

Have you managed to add this feature? Its realy needed.

Thanks
Sergiu Plotnicu

PR #291, opened 2019-10-03
Add the actual Object metadata when add_object_metadata is enabled

hlx98007:
Fix issue #290

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -261,7 +261,8 @@ module Fluent::Plugin
       raw_key = s3["object"]["key"]
       key = CGI.unescape(raw_key)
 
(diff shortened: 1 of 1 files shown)


ganmacs:
Could you check CI?

hlx98007:
@ganmacs Could you guide me on how to resolve this one? I am not too familiar with Ruby and how the CI was setup.

Since the CI test is mocking the data responses and there is no object metadata put initially in the test, I am not sure how to work on the CI part.

repeatedly:
BTW, need new parameter because changing generated record may break existing users.

hlx98007:
@repeatedly If that's the case, I will close the PR because I think this need further discussion.

sppwf:
Hi Guys,

Is there any work done over here?

Thanks
Sergiu Plotnicu

PR #297, opened 2019-10-31
Add support for ECSCredentials to in_s3

### LINKED ISSUE #283: Instance profile credentials not working with ECS
Using the following deps:

```
Successfully installed jmespath-1.4.0
Successfully installed aws-partitions-1.195.0
Successfully installed aws-eventstream-1.0.3
Successfully installed aws-sigv4-1.1.0
Successfully installed aws-sdk-core-3.61.2
Successfully installed aws-sdk-kms-1.24.0
Successfully installed aws-sdk-s3-1.46.0
Successfully installed aws-sdk-sqs-1.20.0
Successfully installed fluent-plugin-s3-1.1.11
```

and the config:
```
    <instance_profile_credentials>
    </instance_profile_credentials>
```

results in:
```
0 unexpected error error_class=Aws::Errors::MissingCredentialsError error="unable to sign request without credentials set"
```

This configuration worked under EC2. The same IAM role applied to the ECS task results in this error. Are there additional IAM roles or configuration required for ECS?

atkinsj:
use ECSCredentials instead of InstanceProfileCredentials if AWS_CONTAINER_CREDENTIALS_RELATIVE_URI is set

fixes #283

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -218,7 +218,7 @@ module Fluent::Plugin
         c = @shared_credentials
         credentials_options[:path] = c.path if c.path
         credentials_options[:profile_name] = c.profile_name if c.profile_name
-        options[:credentials] = Aws::ECSCredentials.new(credentials_options)
+        options[:credentials] = Aws::SharedCredentials.new(credentials_options)
       else
         # Use default credentials
         # See http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -209,12 +209,20 @@ module Fluent::Plugin
         credentials_options[:port] = c.port if c.port
         credentials_options[:http_open_timeout] = c.http_open_timeout if c.http_open_timeout
         credentials_options[:http_read_timeout] = c.http_read_timeout if c.http_read_timeout
-        options[:credentials] = Aws::InstanceProfileCredentials.new(credentials_options)
+        if ENV["AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"]
+          options[:credentials] = Aws::ECSCredentials.new(credentials_options)
+        else
+          options[:credentials] = Aws::InstanceProfileCredentials.new(credentials_options)
+        end
       when @shared_credentials
         c = @shared_credentials
         credentials_options[:path] = c.path if c.path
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_s3.rb
```diff
@@ -209,12 +209,20 @@ def setup_credentials
...
+        end
       when @shared_credentials
         c = @shared_credentials
         credentials_options[:path] = c.path if c.path
         credentials_options[:profile_name] = c.profile_name if c.profile_name
-        options[:credentials] = Aws::SharedCredentials.new(credentials_options)
+        if ENV["AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"]
+          options[:credentials] = Aws::ECSCredentials.new(credentials_options)
+        else
+          options[:credentials] = Aws::InstanceProfileCredentials.new(credentials_options)
```
ganmacs: why don't you use `Aws::SharedCredentials`?

atkinsj: I opted to use the same mechanism that is currently used in the out_s3 module for consistency.

https://github.com/fluent/fluent-plugin-s3/blob/master/lib/fluent/plugin/out_s3.rb#L470

ganmacs: https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/InstanceProfileCredentials.html
https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ECSCredentials.html
https://github.com/fluent/fluent-plugin-s3/blob/1c9c7c8d13fd00aaf7307764f383028636300fe2/fluent-plugin-s3.gemspec#L20
https://github.com/fluent/fluent-plugin-s3/blob/1c9c7c8d13fd00aaf7307764f383028636300fe2/lib/fluent/plugin/in_s3.rb#L59-L64

`InstanceProfileCredentials` and `ECSCredentials`'s initialize method does not accept `path` and `path_name`.

repeatedly: How about this point? From the docs, path and path_name are not supported. I'm not sure why this code works...

atkinsj: Fair point; I don't think that line is necessary at all, I'll push a fix.

atkinsj:
Pushed a fix to stick to SharedCredentials where it should be used and correctly updated the title to reflect the fact that this PR introduces support for ECSCredentials instead of just InstanceProfileCredentials.

Issue #301, opened 2019-11-07
Add Support for IAM Roles for Service accounts (IRSA)


mukgupta:
AWS recently added a new way of authenticating to AWS APIs for Kubernetes Pods called IRSA. For this to work, the aws sdk being used needs to support two environment variables called `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN`. The support for the same has been added in Ruby SDK version`3.65.0` as mentioned [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/). `fluent-plugin-s3` then needs to add support for passing these variables in the dsl

PR #305, opened 2020-01-10
update S3 SDK + fix region access in tests

bonifaido:
Signed-off-by: Nandor Kracser <bonifaido@gmail.com>

This allows us to use the most recent AWS S3 and STS Ruby SDKs and fixes the build as well.

Fixes: https://github.com/fluent/fluent-plugin-s3/issues/301


dnascimento:
Would this support regional sts endpoint?

bonifaido:
This only upgrades the AWS SDK and fixes a test issue which is caused by the update.

ganmacs:
thanks!

PR #307, opened 2020-02-06
First pass at adding AssumeRoleWebIdentity creds

mothershipper:
We wanted to run this plugin on EKS. Instead of acting like kiam or
kube2iam, EKS creates a web token identity file within a container
that allows the application to assume roles -- but using the newer
credential provider.

This commit adds a web_identity_credentials config section and handler
so we can use the identity file within EKS.

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -50,7 +50,7 @@ module Fluent::Plugin
       desc "An identifier for the assumed role session"
       config_param :role_session_name, :string #required
       desc "The absolute path to the file on disk containing the OIDC token"
-      config_param :web_identity_token_file, :string, default: nil #required
+      config_param :web_identity_token_file, :string #required
       desc "An IAM policy in JSON format"
       config_param :policy, :string, default: nil
       desc "The duration, in seconds, of the role session (900-43200)"
@@ -222,8 +222,6 @@ module Fluent::Plugin
         credentials_options[:role_session_name] = c.role_session_name
         credentials_options[:web_identity_token_file] = c.web_identity_token_file
         credentials_options[:policy] = c.policy if c.policy
-        # TODO: Validation or clamping? If we go above 12 hours, the request will fail...
-        #       Is it better behavior to hard fail, or just provide the max session duration?
         credentials_options[:duration_seconds] = c.duration_seconds if c.duration_seconds
         if @s3_region
           credentials_options[:client] = Aws::STS::Client.new(:region => @s3_region)
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -42,6 +42,20 @@ module Fluent::Plugin
       desc "A unique identifier that is used by third parties when assuming roles in their customers' accounts."
       config_param :external_id, :string, default: nil
     end
+    # See the following link for additional params that could be added:
+    # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/STS/Client.html#assume_role_with_web_identity-instance_method
+    config_section :web_identity_credentials, multi: false do
+      desc "The Amazon Resource Name (ARN) of the role to assume"
+      config_param :role_arn, :string # required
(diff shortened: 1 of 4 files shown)


mothershipper:
We're testing this in our EKS cluster today.  I still want to verify the new config works before this is ready to go :)

mothershipper:
Ok, confirmed, this works in our clusters

lib/fluent/plugin/in_s3.rb
```diff
@@ -42,6 +42,20 @@ def initialize
       desc "A unique identifier that is used by third parties when assuming roles in their customers' accounts."
       config_param :external_id, :string, default: nil
     end
+    # See the following link for additional params that could be added:
+    # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/STS/Client.html#assume_role_with_web_identity-instance_method
+    config_section :web_identity_credentials, multi: false do
+      desc "The Amazon Resource Name (ARN) of the role to assume"
+      config_param :role_arn, :string # required
```
ganmacs: You can use `required: true` here like https://github.com/fluent/fluent-plugin-s3/pull/307/files#diff-f4e506e1e7bca94692f1db4ae6dd4716R94.

lib/fluent/plugin/in_s3.rb
```diff
@@ -202,6 +216,19 @@ def setup_credentials
...
         end
         options[:credentials] = Aws::AssumeRoleCredentials.new(credentials_options)
+      when @web_identity_credentials
+        c = @web_identity_credentials
+        credentials_options[:role_arn] = c.role_arn
+        credentials_options[:role_session_name] = c.role_session_name
+        credentials_options[:web_identity_token_file] = c.web_identity_token_file
+        credentials_options[:policy] = c.policy if c.policy
+        # TODO: Validation or clamping? If we go above 12 hours, the request will fail...
+        #       Is it better behavior to hard fail, or just provide the max session duration?
```
ganmacs: How about putting the warning? I think that validation about AWS parameter should be done by AWS, fluentd shouldn't have the knowledge about it.

repeatedly:
Proposal is good. After apply ganmacs reviews, we will merge and release new version.

lib/fluent/plugin/in_s3.rb
mothershipper: Are you sure? The fluentd docs suggest the lack of `default` indicates required for the `config_param` bits.

https://docs.fluentd.org/plugin-development/api-config-types

Excerpt:

```ruby
# Required parameter. The configuration must have this parameter like 'param1 10'
config_param :param1, :integer
# Optional parameter. If the configuration doesn't have 'param2', 100 is used
config_param :param2, :integer, default: 100
```

Also doesn't seem to be reflected upstream: 

https://github.com/fluent/fluentd/blob/d9533da0b41fc652c51d86bf773db6a0075d7973/lib/fluent/config/configure_proxy.rb#L270-L283

That said, I think I set a default for the required value `web_identity_token_file`. I'll push an update in a few.

lib/fluent/plugin/in_s3.rb
mothershipper: Sorry, I'm pretty new to ruby and fluentd -- are you talking about ruby's standard [warning lib](https://ruby-doc.org/core-2.5.0/Warning.html)?

That said, I think you're right. It definitely makes more sense to let AWS validate the parameters. If they ever change the bounds, then we'd have to update the validation (including a warning). I'd personally lean towards not adding the warning in code, and just leaving it in the docs. Happy to add it though if you feel strongly about it!

mothershipper:
Thanks @ganmacs and @repeatedly for the review, I really appreciate the quick response!

I pushed another commit with what I think will address your concerns. If it looks good, I can squash, or you're welcome to squash and merge.

lib/fluent/plugin/in_s3.rb
mothershipper: Oh, I see now you're talking about `log.warn` right?

repeatedly:
sorry for the late. Just merged. I will release new version soon.

PR #308, opened 2020-02-20
Adding support for network proxies in aws assume role config s…

riquemon:
…ection


lib/fluent/plugin/out_s3.rb
```diff
@@ -470,8 +474,19 @@ def setup_credentials
         credentials_options[:policy] = c.policy if c.policy
         credentials_options[:duration_seconds] = c.duration_seconds if c.duration_seconds
         credentials_options[:external_id] = c.external_id if c.external_id
-        if @s3_region
-          credentials_options[:client] = Aws::STS::Client.new(region: @s3_region)
+        credentials_options[:sts_endpoint_url] = c.sts_endpoint_url if c.sts_endpoint_url
```
ganmacs: What's this? Aws::AssumeRoleCredentials.new seems not to support the parameter.
https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleCredentials.html

lib/fluent/plugin/out_s3.rb
```diff
@@ -470,8 +474,19 @@ def setup_credentials
...
+        elsif @region and c.sts_http_proxy
+          credentials_options[:client] = Aws::STS::Client.new(region: @region, http_proxy: c.sts_http_proxy)
+        elsif @region and c.sts_endpoint_url
+          credentials_options[:client] = Aws::STS::Client.new(region: @region, endpoint: c.sts_endpoint_url)
+        elsif c.sts_http_proxy
+          credentials_options[:client] = Aws::STS::Client.new(http_proxy: c.sts_http_proxy)
+        elsif c.sts_endpoint_url
+          credentials_options[:client] = Aws::STS::Client.new(endpoint: c.sts_endpoint_url)
+        elsif @region
+          credentials_options[:client] = Aws::STS::Client.new(region: @region)
```
ganmacs: How about this?

```suggestion
opt = @s3_region ? { region: @s3_region } : {}
opt[:http_proxy] = c.sts_http_proxy if c.sts_http_proxy
opt[:endpoint_url] = c.sts_endpoint_url if c.sts_endpoint_url

credentials_options[:client] = Aws::STS::Client.new(**opt)     
```

ganmacs:
Can you please follow the DCO? https://github.com/fluent/fluent-plugin-s3/pull/308/checks?check_run_id=457118617

lib/fluent/plugin/out_s3.rb
riquemon: We can see in the constructor source code for Aws::AssumeRoleCredentials.new that it accepts a customizable set of options parameters. We can also see that if the option is not in the set of assume_role_options then it will go ahead and set those options to the STS::Clinet constructor. this effectively forwards the STS client parameters for us.

The underlying AWS client and API for STS will take care of the rest for us.

for more reference please refer to awslabs solution in https://github.com/awslabs/aws-fluent-plugin-kinesis/blob/47ad650a907c6524be323b66860e4757fb02ddd9/lib/fluent/plugin/kinesis_helper/client.rb#L117

csolidum:
Is there a plan to get this branch moving forward again? I'm also interested in being able to set sts endpoints for my fluentd deployment.

riquemon:
travis-ci testing seems to be failing due to sts_endpoint_url not being defined in the test config.

I'm not sure why this exception gets raised sense all the clauses where sts_enpoint_url get references are conditioned on the existence of the variable sts_endpoing_url.

please feel free to reach out and let me know of any necessary changes.

lib/fluent/plugin/out_s3.rb
ganmacs: https://github.com/awslabs/aws-fluent-plugin-kinesis/blob/47ad650a907c6524be323b66860e4757fb02ddd9/lib/fluent/plugin/kinesis_helper/client.rb#L47

awslabs defines `sts_endpoint_url` here so it can access it. but you don't. it causes CI fail.
Maybe you should define in `assume_role_credentials`  not in `web_identity_credentials ` .
https://github.com/fluent/fluent-plugin-s3/blob/b3210b791f1b4df36853986a8e1068f40fe61f9f/lib/fluent/plugin/out_s3.rb#L31-L44

riquemon:
Resolved conflicts with master branch and fixed issue that cause test cause failure.
This update should now allow the config file to set an http proxy for assume role credentials.

This feature is to support fluent installations running behind an AWS network proxy.
please review

lib/fluent/plugin/out_s3.rb
```diff
@@ -41,6 +41,10 @@ def initialize
       config_param :external_id, :string, default: nil, secret: true
       desc "The region of the STS endpoint to use."
       config_param :sts_region, :string, default: nil
+      desc "A http proxy url for requests to aws sts service"
+      config_param :sts_http_proxy, :string, default: nil, secret: true
+      desc "A url for a regional sts api endpoint, the default is global"
+      config_param :sts_endpoint_url, :string, default: nil ## STS_region
```
ganmacs, line 47: What does it mean? > `## STS_region`

lib/fluent/plugin/out_s3.rb
```diff
@@ -489,7 +493,19 @@ def setup_credentials
         credentials_options[:policy] = c.policy if c.policy
         credentials_options[:duration_seconds] = c.duration_seconds if c.duration_seconds
         credentials_options[:external_id] = c.external_id if c.external_id
-        if c.sts_region
+        credentials_options[:sts_endpoint_url] = c.sts_endpoint_url if c.sts_endpoint_url
+        credentials_options[:sts_http_proxy] = c.sts_http_proxy if c.sts_http_proxy
+        if c.sts_http_proxy and c.sts_endpoint_url
```
ganmacs, line 498: Please use `&&` instead of `and` because they have different precedence. It's confusing.

https://ruby-doc.org/core-2.6.2/doc/syntax/precedence_rdoc.html#top

lib/fluent/plugin/out_s3.rb
```diff
@@ -489,7 +493,19 @@ def setup_credentials
...
+          credentials_options[:client] = Aws::STS::Client.new(http_proxy: c.sts_http_proxy, endpoint: c.sts_endpoint_url)
+        elsif @region and c.sts_http_proxy
```
ganmacs, line 500: same here https://github.com/fluent/fluent-plugin-s3/pull/308/files#r449343157

lib/fluent/plugin/out_s3.rb
```diff
@@ -489,7 +493,19 @@ def setup_credentials
...
+          credentials_options[:client] = Aws::STS::Client.new(region: @region, http_proxy: c.sts_http_proxy)
+        elsif @region and c.sts_endpoint_url
```
ganmacs, line 502: same here https://github.com/fluent/fluent-plugin-s3/pull/308/files#r449343157

repeatedly:
Looks good to me. After apply ganmacs review, we will merge and release new version.

lib/fluent/plugin/out_s3.rb
riquemon: just a comment, it wrote it there to help me figure out what everything means.
I'll remove it with my next push.

lib/fluent/plugin/out_s3.rb
riquemon: sounds good, I'll update to use `&&` instead.

repeatedly:
Thanks!

PR #309, opened 2020-02-25
Add SNS event support

tabletcorry:
Add support for SQS messages arriving via SNS. This is commonly used to
fan out S3 events.

Because SNS wraps the messages, we need to dive one layer deeper into
the SQS body.

https://aws.amazon.com/blogs/compute/fanout-s3-event-notifications-to-multiple-endpoints/

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -104,6 +104,8 @@ module Fluent::Plugin
       config_param :wait_time_seconds, :integer, default: 20
       desc "Polling error retry interval."
       config_param :retry_error_interval, :integer, default: 300
+      desc "Assume messages as SNS notifications"
+      config_param :assume_sns, :bool, default: false
     end
 
     desc "Tag string"
@@ -182,6 +184,10 @@ module Fluent::Plugin
           begin
             body = Yajl.load(message.body)
             log.debug(body)
+            if @assume_sns
+              next unless body["Message"]
+              body = Yajl.load(body["Message"])
+            end
             next unless body["Records"] # skip test queue
 
             process(body)


lib/fluent/plugin/in_s3.rb
```diff
@@ -104,6 +104,8 @@ def initialize
       config_param :wait_time_seconds, :integer, default: 20
       desc "Polling error retry interval."
       config_param :retry_error_interval, :integer, default: 300
+      desc "Assume SNS fanout"
```
ganmacs: "Assume messages as SNS notifications" is better for me.

lib/fluent/plugin/in_s3.rb
```diff
@@ -181,6 +183,11 @@ def run
         @poller.poll(options) do |message|
           begin
             body = Yajl.load(message.body)
+            if @assume_sns
+              log.debug(body)
```
ganmacs: how about putting the following after L191? it can handle both SNS messages and other normal messages.

```rb
if @assume_sns && body['Message']
  body = Yajl.load(body['Message'])
end
```

tabletcorry: Hmm, I guess? Seems like an optimization for an impossible case, where a single bucket is sending events via both SNS and SQS. This would be an extremely unusual situation I think?

ganmacs: Ah, You're right. it's okay as is.

lib/fluent/plugin/in_s3.rb
```diff
@@ -181,6 +183,11 @@ def run
...
+              next unless body["Message"]
+              body = Yajl.load(body["Message"])
+            end
```
ganmacs: How about this? Besides, Can you delete `log.debug(body)` at L184? it's duplicated.
```suggestion
            log.debug(body)
            if @assume_sns
              next unless body["Message"]
              body = Yajl.load(body["Message"])
            end
```

ganmacs:
Can you follow the DCO? https://github.com/fluent/fluent-plugin-s3/pull/309/checks?check_run_id=467960999

ganmacs:
Thank you!!

tabletcorry:
I derped this. The fix is currently in my working branch: https://github.com/tabletcorry/fluent-plugin-s3/commit/32514ba28c4b6f5527b352c91bb77d5b1f391b8b

ganmacs:
Ah, nice catch! looks good. can you create the PR?

PR #316, opened 2020-04-13
Support S3 Dual-Stack Endpoints in output plugin

andreaso:
Only doing this for the output plugin, since there are no
corresponding SQS Dual-Stack Endpoints, making the change less
meaningful for the input plugin.

Signed-off-by: Andreas Olsson <andreas@arrakis.se>


repeatedly:
Thanks!

PR #320, opened 2020-05-05
Update plugin's version in installation guide

Punkoivan:
Hello folks,
I've used that line from docs while installed plugin v 1.0.0 and then spent around 1 hour while figured out why "add_object_metadata" is not used :) 

So I upgraded docs to install version 1.3.0

Thanks


Punkoivan:
Sorry, I did it in a wrong way.

PR #322, opened 2020-05-05
Update plugin's version in installation guide

Punkoivan:
Hello folks,
I've used that line from docs while installed plugin v 1.0.0 and then spent around 1 hour while figured out why "add_object_metadata" is not used :)

So I upgraded docs to install version 1.3.0

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ We must setup SQS queue and S3 event notification before use this plugin.
 Simply use RubyGems:
 
     $ gem install fluent-plugin-s3 -v "~> 0.8" --no-document # for fluentd v0.12 or later
-    $ gem install fluent-plugin-s3 -v 1.0.0 --no-document # for fluentd v1.0 or later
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ We must setup SQS queue and S3 event notification before use this plugin.
 Simply use RubyGems:
 
     $ gem install fluent-plugin-s3 -v "~> 0.8" --no-document # for fluentd v0.12 or later
-    $ gem install fluent-plugin-s3 -v 1.0.0 --no-document # for fluentd v1.0 or later
+    $ gem install fluent-plugin-s3 -v 1.3.0 --no-document # for fluentd v1.0 or later
```
repeatedly, line 35: Hmm... removing `-v 1.3.0` is more better. It installs latest version.

Punkoivan: Agreed. One reason I've chosen 1.3.0 since it seems to be super stable.
If we can guarantee that latest version is the best choice, let's remove version completely.

repeatedly:
Merged. We will consider more better command in the future.

PR #329, opened 2020-05-21
Ignoring errors when we can't assume the role

caiusb:
This will prevent fluentd from crashing all together

This will require elva changes to use the new plugin. Changes are on the way for that one.


caiusb:
Opened in error.

Issue #331, opened 2020-06-03
Special s3_endpoints are possible even in amazonaws.com


nyetsche:
S3 in the public AWS environment has has more than one endpoint, most notably FIPS. Ex:`s3-fips-us-gov-west-1.amazonaws.com`

Ref: https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html

The code at https://github.com/fluent/fluent-plugin-s3/blob/bd225ea951b658f23ac156cd511dfd4aa9927398/lib/fluent/plugin/out_s3.rb#L174 raises an error and fails if you use `s3_endpoint` and the endpoint ends in `amazonaws.com`. This should be removed.

PR #332, opened 2020-06-03
remove s3_endpoint check for amazonaws.com domain.

nyetsche:
Multiple endpoints may exist, notably for FIPS S3.

The FIPS SQS endpoint is not different from the usual one, but it's
removed for consistency.

https://github.com/fluent/fluent-plugin-s3/issues/331

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -118,14 +118,6 @@ module Fluent::Plugin
     def configure(conf)
       super
 
-      if @s3_endpoint && @s3_endpoint.end_with?('amazonaws.com')
-        raise Fluent::ConfigError, "s3_endpoint parameter is not supported for S3, use s3_region instead. This parameter is for S3 compatible services"
-      end
-
-      if @sqs.endpoint && @sqs.endpoint.end_with?('amazonaws.com')
-        raise Fluent::ConfigError, "sqs/endpoint parameter is not supported for SQS, use s3_region instead. This parameter is for SQS compatible services"
-      end
-
       parser_config = conf.elements("parse").first
       unless @sqs.queue_name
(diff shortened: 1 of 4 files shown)


repeatedly:
I checked the document. Do these endpoints have `-gov-`?
If so, could you check `-gov-` for validation skip. 
We want to keep validation for normal S3 users.

In addition, could you follow DCO?

https://github.com/apps/dco

nyetsche:
No, the endpoints are not always with `-gov-`. https://aws.amazon.com/compliance/fips/ has a list - ex: `s3-fips.us-east-1.amazonaws.com` is valid. (Although you need to use the domain style, so only `BUCKETNAME.s3-fips.us-east-1.amazonaws.com` will resolve).

repeatedly:
I see. So changing conditino to  `if @s3_endpoint && (@s3_endpoint.end_with?('amazonaws.com') && !['fips', 'gov'].any? { |e| @s3_endpoint.include?(e) })` is better.

jared-gs:
Could this check just be removed entirely? The feel the discretion to use `s3_endpoint` is just best left to user. Putting in `fips` and `gov` solves the immediate problem, but not future edge-cases.

repeatedly:
> Could this check just be removed entirely?

AWS makes several parameters deprecated. So I can't say these options are safe for users.

https://aws.amazon.com/jp/blogs/developer/working-with-regions/

nyetsche:
@repeatedly  - I've created a new PR with your suggestions at https://github.com/fluent/fluent-plugin-s3/pull/335

I'm not familiar with this Ruby testing framework - why are the SQS checks failing? I've merely changed the `if` statement to allow for `fips` or `gov`, as you mentioned.

repeatedly:
Closed via https://github.com/fluent/fluent-plugin-s3/pull/335
Thanks for the work 👍

PR #336, opened 2020-06-24
Add ability to set sts region

csolidum:
This would be useful for fluentd deployments running in a different region than the S3 bucket it's writing to, but wants to take advantage of the STS endpoint. In the case this is not set, falls back to the region of the s3 bucket


repeatedly:
Looks good.

Issue #343, opened 2020-07-28
No detail error information in the plugin logs.


henrysher:
I have met an issue in "SQS Polling Failed." but no idea what happened even I have enabled the debug logs.
Then I noticed that in the code "lib/fluent/plugin/in_s3.rb".

It does capture the exceptions but does not to record it in the plugin logs.
After I added the code to record the exception, i got the "access denied" error:

> 2020-07-28 06:04:08 +0000 [warn]: #0  error_class=Aws::SQS::Errors::AccessDenied error="Access to the resource https://sqs..../4233XXXXXXXXX/elbaccess-XXXXXXXXXXXXXX is denied."

```ruby
    def run
      options = {}
      options[:wait_time_seconds] = @sqs.wait_time_seconds
      options[:skip_delete] = @sqs.skip_delete
      @poller.before_request do |stats|
        throw :stop_polling unless @running
      end
      begin
        @poller.poll(options) do |message|
          begin
            body = Yajl.load(message.body)
            log.debug(body)
            next unless body["Records"] # skip test queue

            process(body)
          rescue => e
            log.warn(error: e)
            log.warn_backtrace(e.backtrace)
            throw :skip_delete
          end
        end
      rescue => e
        log.warn("SQS Polling Failed. Retry in #{@sqs.retry_error_interval} seconds")
        # added by Henry -  exception captured but no recorded in the logs.
        sleep(@sqs.retry_error_interval)
        retry
      end
    end
```

henrysher:
Thank!

PR #344, opened 2020-07-28
Fix the #343 to add more logs for captured exceptions

henrysher:
Fix the issue #343

---
This change is [Reviewable](https://reviewable.io/reviews/fluent/fluent-plugin-s3/344)

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -192,7 +192,7 @@ module Fluent::Plugin
           end
         end
       rescue => e
-        log.warn("SQS Polling Failed. Retry in #{@sqs.retry_error_interval} seconds")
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_s3.rb
```diff
@@ -193,6 +193,7 @@ def run
         end
       rescue => e
         log.warn("SQS Polling Failed. Retry in #{@sqs.retry_error_interval} seconds")
+        log.warn(error: e)
```
repeatedly, line 196: How about merging this line into previous `log.warn`?

```
log.warn("SQS Polling  ...  seconds", error: e)
```

repeatedly:
In addition, could you add sign to your commit? > https://github.com/apps/dco
You can use `--amend --signoff`.

henrysher:
Please check the PR #345

Issue #160, opened 2016-07-29
S3 cost optimisation: Remember the last index value that was used


tomfitzhenry:
## The problem
Suppose you use the defaults:
* `s3_object_key_format` of `%{path}%{time_slice}_%{index}.%{file_extension}` 
* `time_slice_format` of `%Y%m%d%H`

And suppose you flush every 30 seconds. So 120 files per hour.

The first file will check whether `foo-2016010109_1.gz` exists via an S3 HEAD request, see it doesn't exist, and then upload to that filename.

The next file will be first check whether `foo-2016010109_1.gz` exists via an S3 HEAD request, see it exists, and so increment the index to `foo-2016010109_2.gz`, check whether it exists via  with an S3 HEAD request, see it doesn't exist, and then upload to the filename.

This will continue. When we get to the final file of the hour (the 120th file), we'll first do 119 HEAD requests!

That's 1+2+...+119 = 7140 S3 requests over the hour.

S3 HEAD requests are "$0.004 per 10,000 requests". So the monthly cost of the above, for 5 log files for 100 instances amounts to 7140*5*100*24*30*$0.004/1000 = $1028

And that's per input log file, per instance.

More generally, 1+2+...+n is O(n^2) and we can reduce this to O(n).

## Solutions
(a) The user can modify `time_slice_format` to include `%M`. Or the default could include `%M`.
(b) fluent-plugin-s3 could remember the last index it uploaded to, and so not have to check whether they already exist: fluentd would know they do.

If either solution was implemented, we'd've reduced the number of HEAD requests from O(n^2) to O(n).

So rather than 7140 S3 requests per hour per log file per instance, we'd only do 120.

This reduces the monthly cost from $1028 to $17.

cnorthwood:
The issue with (b) is if you have multiple instances writing to the same path, you'd still need to have the collision check.

tomfitzhenry:
Oops. I updated (b) to clarify that the HEAD request would still occur for the index we're trying, but it then wouldn't need to try the n-1 earlier requests.

If we're up to index n, we needn't check 1, 2, ..., n-1. But fluentd-plugin-s3 seems to: See https://github.com/fluent/fluent-plugin-s3/blob/master/lib/fluent/plugin/out_s3.rb#L190 and line 219.

bbc-tomfitzhenry:
[...]

PR #353, opened 2020-10-02
add warning to readme

caleb15:
see https://github.com/fluent/fluent-plugin-s3/issues/160

The bug has cost people thousands of dollars and until fixed it would be beneficial for people to be extra aware of it.


caleb15:
unstale

kenhys (approved):
It seems reasonable until the issue is resolved.

Issue #326, opened 2020-05-12
process of determining 'index' is not thread safe


tatsu-yam:
* fluentd or td-agent version : td-agent 3.4.1-0 (fluentd 1.4.2)
* fluent-plugin-s3 version    : 1.3.1
* Your problem explanation. If you have an error logs, write it together.

In fluent-plugin-s3, if the value of flush_thread_count is greater than 1 , the data on S3 will be missing. I think that it is because `Fluent::Plugin::S3Output#write` method is not thread safe.

td-agent.conf
```
<source>
  @type tail
  format ltsv
  keep_time_key true
  read_from_head true
  path /tmp/td-agent-failure-sample/tmp/1500000.log
  pos_file /var/log/td-agent/1500000.log-ng.pos
  tag app.test_log
  time_format %d/%b/%Y:%H:%M:%S %z
</source>

<match app.**>
  @type s3
  @id out_s3

  s3_bucket "#{ENV['S3_BUCKET']}"
  s3_region "#{ENV['S3_REGION']}"

  path "20200512_1647/ng"
  s3_object_key_format "%{path}/%Y%m%d.%{index}.dat.%{file_extension}"

  check_apikey_on_start false
  check_bucket false

  <buffer tag,time>
    @type file
    path /var/log/td-agent/buffer/ng
    timekey 1d
    timekey_wait 10m
    chunk_limit_size 256m
    flush_thread_count 4

    flush_at_shutdown true
    flush_mode interval
    flush_interval 5 # test setting
  </buffer>
  <format>
    @type json
  </format>
</match>
```

Test data (`/tmp/td-agent-failure-sample/tmp/1500000.log`) was transferred to s3 with td-agent.
As a result, the number of records in the original data and the data on s3 do not match. Of course, there are no errors in td-agent.log.
```
# source data
$ cat /tmp/td-agent-failure-sample/tmp/1500000.log |wc -l
1500000

# s3
$ aws s3 sync s3://$S3_BUCKET/20200512_1647/ ans/20200512_1647/
$ zcat ans/20200512_1647/ng/* |wc -l
1156632
```

So, I made the following modifications to fluentd and fluent-plugin-s3 and transferred it.
```diff
# fluentd code diff
$ git diff
diff --git a/lib/fluent/log.rb b/lib/fluent/log.rb
index 3424aa9..76c09c1 100644
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
[...]

PR #355, opened 2020-10-24
Update index safely between threads. Fixes #160 and #326.

### LINKED ISSUE #160: S3 cost optimisation: Remember the last index value that was used

worr:
As mentioned in a warning, as well as https://github.com/fluent/fluent-plugin-s3/issues/326
and https://github.com/fluent/fluent-plugin-s3/issues/160, the process of
determining the index added to the default object key is not thread-safe. This
adds some thread-safety until version 2.x is out where chunk_id is used
instead of an index value.

This is not a perfect implementation, since there can still be races between
different workers if workers are enabled in fluentd, or if there are multiple
fluentd instances uploading to the same bucket. This commit is just to resolve
this problem short-term in a way that's backwards compatible.

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -2,6 +2,7 @@ require 'fluent/plugin/output'
 require 'fluent/log-ext'
 require 'fluent/timezone'
 require 'aws-sdk-s3'
+require 'concurrent'
 require 'zlib'
 require 'time'
 require 'tempfile'
@@ -223,7 +224,7 @@ module Fluent::Plugin
       # For backward compatibility
       # TODO: Remove time_slice_format when end of support compat_parameters
       @configured_time_slice_format = conf['time_slice_format']
-      @values_for_s3_object_chunk = {}
+      @values_for_s3_object_chunk = Concurrent::Hash.new
       @time_slice_with_tz = Fluent::Timezone.formatter(@timekey_zone, @configured_time_slice_format || timekey_to_timeformat(@buffer_config['timekey']))
     end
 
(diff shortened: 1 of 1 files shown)


worr:
I tested this by deploying this to a kubernetes cluster with a single fluentd, and generating artificial log load. This uses substantially fewer `HEAD` calls than the existing code, and because of that, can upload faster, especially in pathological cases.

concurrent-ruby will use C-based atomic primitives when available, or else it will fall back on a simple mutex.

Again, this is not a bullet-proof solution, since the default object name can still collide with multiple fluentd's or multiple workers.

PR #361, opened 2020-12-17
Bucket name from record field

rmontenegroo:
Changed the plugin so as bucket name can be set dynamically by a record field. 

A new parameter s3_bucket_key may be set informing plugin to set the bucket name from record['key'].

If s3_bucket_key is not found in chunk keys or key is not found in record, s3_bucket is used as a fallback.


cosmo0920 (changes requested):
We should use [built-in placeholders](https://docs.fluentd.org/configuration/buffer-section#placeholders) instead of using xxx_key parameter and manipulating record.
This patch uses old Fluentd record manipulation style. This doesn't fit Fluentd v1 world philosophy.

rmontenegroo:
Making another PR...

PR #363, opened 2020-12-19
Bucket name from placeholder

rmontenegroo:
This path allows the use of placeholders to setup bucket names.

There is a new parameter s3_bucket_fallback to use in cases when at least one placeholder is not found amoung chunk keys.


lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +277,35 @@ def write(chunk)
                    else
                      @time_slice_with_tz.call(metadata.timekey)
                    end
+      
+      bucket_name = nil
+
+      if @s3_bucket =~ /\$\{.*\}/
+        @s3_bucket.scan(/\$\{([^\$\{\}]+)\}/) do |placeholder|
```
cosmo0920, line 284: Time placeholders does not use $ character. 
We can use Validator class instead.
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L607

rmontenegroo:
@cosmo0920 why get_placeholders_time does not return any time placeholder if I set @s3_bucket to "test-%Y" for example?
If I set @s3_bucket = "test-%Y-%m-%d" get_placeholders_time return only %d.

rmontenegroo:
The thing is PlaceholderValidator does not allow a timekey larger than a day (https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L686) so it won't be possible to validate time placeholders for year and month using the validator. Am I right? Something like "test-%Y-%m" won't be accepted by validator as placeholders.

rmontenegroo:
@cosmo0920 I implemented a simple ext_get_placeholders_time only to check if there are time markers (%Y %m %d %H %M %S) in a string. With this I am able to check time placeholders, and with Validator class I can check tag and keys placeholders. I hope it is satisfying.

cosmo0920:
> The thing is PlaceholderValidator does not allow a timekey larger than a day (https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L686) so it won't be possible to validate time placeholders for year and month using the validator. Am I right?

Right.

test/test_out_s3.rb
```diff
@@ -427,7 +427,7 @@ def setup_mocks(exists_return = false)
     mock(Aws::S3::Client).new(anything).at_least(0) { @s3_client }
     @s3_resource = mock(Aws::S3::Resource.new(client: @s3_client))
     mock(Aws::S3::Resource).new(client: @s3_client) { @s3_resource }
-    @s3_bucket = mock(Aws::S3::Bucket.new(name: "test",
+    @s3_bucket = mock(Aws::S3::Bucket.new(name: "test_bucket",
```
cosmo0920, line 430: Why this change is needed?
To get merged this PR, we should add placeholders testcases.

cosmo0920 (changes requested):

test/test_out_s3.rb
rmontenegroo: I sincerely don't know. I thought someone changed it last patch. It was producing an error on tests I guess. I will check it later.

Yeah... I was afraid you would say that about testing placeholders. I truly have no idea how to do it right now since they're evaluated during plug-in execution time.

Any suggestions? If someone give a hint I might get through it. I am no expert on ruby tests though.

lib/fluent/plugin/out_s3.rb
cosmo0920:
```ruby
if !@tag_placeholders.empty? || !@keys_placeholders.empty? || !@time_placeholders.empty?
```
?

rmontenegroo: Tag, time and keys placeholders are scanned on @s3_bucket on star. If they're empty, the bucket is created during the star and anytime else in the future.

On the other way, if placeholders are found in @s3_bucket, it means bucket object must be evaluated, everything time before a buffer flush. This is done inside the write method. In the write method I check this situation evaluating whether @bucket is instantiated or not.

test/test_out_s3.rb
cosmo0920: > Yeah... I was afraid you would say that about testing placeholders. I truly have no idea how to do it right now since they're evaluated during plug-in execution time.

These testcases are fluent-plugin-elasticsearch's but they shows how to write placeholders testcases:
https://github.com/uken/fluent-plugin-elasticsearch/blob/c0a4713edfc620cf566ede02f0f85930d1c5a88a/test/plugin/test_out_elasticsearch.rb#L3893

* Write configuration which includes `<buffer attributes1, attributes2, ..., attributesN>` with `Fluent::Config::Element.new` and using tag/time/user defined(variables) placeholders
* Creates chunk metadata keys which are originated from `<buffer attributes1, attributes2, ..., attributesN>` (This is automatically created when use test driver#feed)
* Assert placeholders replaced values with test-unit assertions

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+          placeholder = placeholder.join
```
cosmo0920, line 306:
```ruby
placeholders = placeholder.join
```

placeholder variable should contain multiple placeholder**s** information?
If so, this variable name should be plural form.

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+          if (not chunk.metadata.variables) or (not chunk.metadata.variables.keys.include?(placeholder.to_sym))
```
cosmo0920, line 307:
```ruby
if (!chunk.metadata.variables) || (!chunk.metadata.variables.keys.include?(placeholder.to_sym))
```

For unifying coding style.

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+        bucket = @bucket
+      end
+
+      bucket = @bucket ? @bucket : nil
+
+      if (not bucket) and (not @tag_placeholders.empty?)
+        if (not chunk.metadata.tag) or ((@tag_placeholders.max + 1) > chunk.metadata.tag.split('.').length)
+          bucket = create_bucket(use_fallback("tag"))
+        end
+      end
```
cosmo0920, line 337:
```ruby
      if !bucket && !@tag_placeholders.empty?
        if !chunk.metadata.tag || ((@tag_placeholders.max + 1) > chunk.metadata.tag.split('.').length)
          bucket = create_bucket(use_fallback("tag"))
        end
      end
```

For unifying coding style, too.

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+
+      if not bucket
```
cosmo0920, line 339:
```ruby
if !bucket
```

or

```ruby
unless bucket
```

is better.

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+        @keys_placeholders.each do |placeholder|
+          if (not chunk.metadata.variables) or (not chunk.metadata.variables.keys.include?(placeholder.to_sym))
```
cosmo0920, line 341:
```ruby
          if !chunk.metadata.variables || !chunk.metadata.variables.keys.include?(placeholder.to_sym)
```

is better.

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+            bucket = create_bucket(use_fallback(placeholder))
+            break
+          end
+        end
+      end
+
+      if (not bucket) and (not chunk.metadata.timekey) and @time_placeholders
```
cosmo0920, line 348:
```ruby
      if !bucket && !chunk.metadata.timekey && @time_placeholders
```

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+        bucket = create_bucket(use_fallback("time"))
+      end
+
+      if not bucket
```
cosmo0920, line 352:
```ruby
if !bucket
```

or

```ruby
unless bucket
```

is better, too

lib/fluent/plugin/out_s3.rb
```diff
@@ -272,6 +298,60 @@ def write(chunk)
...
+              bucket_name = @s3_bucket_fallback
+              log.warn "Using @s3_bucket_fallback ('#{@s3_bucket_fallback}') as a fallback bucket name."
+              break
+            else
+              raise "It was possible to extract placeholder '#{placeholder}' from chunk and @s3_bucket_fallback is not set."
+            end
+          end
+        end
+
+        if not bucket_name
```
cosmo0920, line 319:
```ruby
if !bucket_name
```

or

```ruby
unless bucket_name
```

is better.

cosmo0920 (changes requested):
We use `!`, `&&`, `||` instead of `not`, `and`, and `or` operator.
This is because symbol versions are higher precedence than `not`, `and`, and `or` operators. This is some pitfall of Ruby.
https://ruby-doc.org/core-2.4.0/doc/syntax/precedence_rdoc.html

PR #364, opened 2020-12-24
Multiple S3 buckets support for input, dynamic bucket names from SQS, and queue url override

abdullahzen:
###Changes made:
- Allowing multiple s3 bucket names in the input s3 plugin and preserving the previous functionality with a single bucket name
- Added dynamic fetching of objects by bucket names returned by SQS. (while checking that the returned bucket name is one of the bucket names that were provided in the s3_buckets and were verified)
- Added queue url to the sqs section to override calling get_queue_url in some cases where get_queue_url is not allowed for specific IAM users.
- Changed the default tag to include the bucket name


abdullahzen:
@cosmo0920 Can I get a review? thank you!

lib/fluent/plugin/in_s3.rb
```diff
@@ -76,8 +76,8 @@ def initialize
       desc "Profile name. Default to 'default' or ENV['AWS_PROFILE']"
       config_param :profile_name, :string, default: nil
     end
-    desc "S3 bucket name"
-    config_param :s3_bucket, :string
+    desc "S3 bucket name(s) separated by commas in case of multiple bucket names"
+    config_param :s3_buckets, :string
```
cosmo0920, line 80: Changing parameter name breaks existing configurations.
We couldn't accept this change.
Instead, using deprecated attr in config_param and adding parameter conversion code are needed to accept this PR.

cosmo0920 (changes requested):

lib/fluent/plugin/in_s3.rb
abdullahzen: Thank you for the review. I'll do the requested changes to make it backwards compatible and get back to you 👍

PR #367, opened 2021-02-25
Enable multi process workers support

asenci:
Enables Fluentd multi process workers support for the input plugin.

Concurrency safety is enforced by the SQS queue, so no additional code is required.

This is specially advantageous for high BDP connections (high latency, high throughput), because it allows multiple simultaneous download streams from S3.


ashie (approved):
Sorry for long silence, it seems fine.

Issue #382, opened 2021-07-28
Need use S3 endpoint in AWS VPC


rubroboletus:
### Describe the bug

In source, you are checking for "fips" and "gov" inside amazonaws.com, when not there, s3_endpoint is rejected with error. Need to enable same functionality for VPCE endpoints.

### To Reproduce

Create AWS VPC, S3 VPCE endpoint, try setting s3_endpoint to it.

### Expected behavior

Use s3_endpoint value, when is ends with vpce.amazonaws.com

### Your Environment

```markdown
- Fluentd version: any
- TD Agent version:
- fluent-plugin-s3 version: any
- aws-sdk-s3 version: any
- aws-sdk-sqs version: any
- Operating system: any
- Kernel version: any
```

### Your Configuration

s3_endpoint bucket.vpce-07a8af356c72bdd58-j7hw3pjl.s3.eu-central-1.vpce.amazonaws.com

### Your Error Log

```shell
s3_endpoint parameter is not supported for S3, use s3_region instead. This parameter is for S3 compatible services
```

kenhys:
Hmm, Is it enough to allow `vpce` in `@s3_endpoint`?

```ruby
diff --git a/lib/fluent/plugin/in_s3.rb b/lib/fluent/plugin/in_s3.rb
index 5c54298..26177c9 100644
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -118,7 +118,7 @@ module Fluent::Plugin
     def configure(conf)
       super
 
-      if @s3_endpoint && (@s3_endpoint.end_with?('amazonaws.com') && !['fips', 'gov'].any? { |e| @s3_endpoint.include?(e) })
+      if @s3_endpoint && (@s3_endpoint.end_with?('amazonaws.com') && !['fips', 'gov', 'vpce'].any? { |e| @s3_endpoint.include?(e) })
         raise Fluent::ConfigError, "s3_endpoint parameter is not supported for S3, use s3_region instead. This parameter is for S3 compatible services"
       end
```

rubroboletus:
I hope, that "vpce" was created just for this VPC Endpoints. But not a AWS employee.

kenhys:
Fixed by #384.

PR #384, opened 2021-08-18
Don't raise error when s3_endpoint is used for VPC endpoint

kenhys:
Usually, using s3_region is recommended, but when accessing
s3 via VPC, it should not raise an error.

VPC endpoint(For example via private link) is assigned to
xxx.vpce.amazonaws.com, so it should be added as an exceptional case.

See https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html

Closes: #382


cosmo0920 (approved):
:+1:

Issue #391, opened 2022-03-08
The default `s3_object_key_format` is racy with multi workers


daipom:
### Is your feature request related to a problem? Please describe.

I assume the same problem could occur in multi-workers as #326, although I have not confirmed the problem.

In #326, multi-threading with `flush_thread_count` is discussed,
but I think the same problem exists when running with multiple workers.

### Describe the solution you'd like

In #327, the warning was added for the case of multi-threading with `flush_thread_count`.
So we need to add the same warning for the case of multi-workers.

### Describe alternatives you've considered

None.

daipom:
#392 is merged, so I close this.

PR #392, opened 2022-03-09
Add warning on multi workers

daipom:
This is a fix for #391.

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -461,7 +461,8 @@ module Fluent::Plugin
         log.warn "The default value of s3_object_key_format will use ${chunk_id} instead of %{index} to avoid object conflict in v2"
       end
 
-      if (@buffer_config.flush_thread_count > 1) && ['${chunk_id}', '%{uuid_flush}'].none? { |key| @s3_object_key_format.include?(key) }
+      is_working_on_multi_thread = @buffer_config.flush_thread_count > 1 || system_config.workers > 1
+      if is_working_on_multi_thread && ['${chunk_id}', '%{uuid_flush}'].none? { |key| @s3_object_key_format.include?(key) }
         log.warn "No ${chunk_id} or %{uuid_flush} in s3_object_key_format with multiple flush threads. Recommend to set ${chunk_id} or %{uuid_flush} to avoid data lost by object conflict"
       end
     end
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -461,9 +461,9 @@ module Fluent::Plugin
         log.warn "The default value of s3_object_key_format will use ${chunk_id} instead of %{index} to avoid object conflict in v2"
       end
 
-      is_working_on_multi_thread = @buffer_config.flush_thread_count > 1 || system_config.workers > 1
-      if is_working_on_multi_thread && ['${chunk_id}', '%{uuid_flush}'].none? { |key| @s3_object_key_format.include?(key) }
-        log.warn "No ${chunk_id} or %{uuid_flush} in s3_object_key_format with multiple flush threads. Recommend to set ${chunk_id} or %{uuid_flush} to avoid data lost by object conflict"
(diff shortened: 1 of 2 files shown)


daipom:
This just fixes a warning log.
I don't have the environment, so I can't check the behavior.
However, I wrote test codes and checked all success of  `configure` subtest.

fujimotos:
I think the general direction of this patch is good. However:

> So we need to add the same warning for the case of multi-workers.

Won't this implementation produce false-positives? For example, if we use multiple workers but put `out_s3` in `<worker 0>`, that should be a perfectly fine way to use this plugin. 

Nonetheless, this patch is likely to produce warnings on such cases, which I think can be confusing to users.

lib/fluent/plugin/out_s3.rb
```diff
@@ -461,7 +461,8 @@ def check_s3_path_safety(conf)
         log.warn "The default value of s3_object_key_format will use ${chunk_id} instead of %{index} to avoid object conflict in v2"
       end
 
-      if (@buffer_config.flush_thread_count > 1) && ['${chunk_id}', '%{uuid_flush}'].none? { |key| @s3_object_key_format.include?(key) }
+      is_working_on_multi_thread = @buffer_config.flush_thread_count > 1 || system_config.workers > 1
+      if is_working_on_multi_thread && ['${chunk_id}', '%{uuid_flush}'].none? { |key| @s3_object_key_format.include?(key) }
         log.warn "No ${chunk_id} or %{uuid_flush} in s3_object_key_format with multiple flush threads. Recommend to set ${chunk_id} or %{uuid_flush} to avoid data lost by object conflict"
```
fujimotos, line 466: I think we can tweak this error message to cover multiple workers,
so that it's clear to users why they got this warning.

In particular, I'm thinking something like this:

> "No ${chunk_id} or %{uuid_flush} in s3_object_key_format with multiple flush threads **(or multiple workers).**
> Recommend to set ${chunk_id} or %{uuid_flush} to avoid data lost by object conflict"

daipom: Thank you!
I have fixed it.

daipom:
> I think the general direction of this patch is good. However:
> 
> > So we need to add the same warning for the case of multi-workers.
> 
> Won't this implementation produce false-positives? For example, if we use multiple workers but put `out_s3` in `<worker 0>`, that should be a perfectly fine way to use this plugin.
> 
> I'm thinking something like this:
> 
> ```
> <system>
>   worker 3
> </system>
> <worker 0>
>   <match **>
>      @type s3
>      ...
>    </match>
> </worker>
> ```
> 
> Nonetheless, this patch is likely to produce warnings on such cases, which I think can be confusing to users.

In this case, `system_config.workers` is `1`, so the warning is not produced.
(https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/base.rb#L61)

We should also have a test code for this case,
but I don't know how to specify the worker directive on the test code, so I can't do that right away.

ashie:
BTW I've fixed CI failure which isn't concerned with this PR in #393

daipom:
@ashie Thank you! I have rebased this branch and all checks have passed.

fujimotos:
Merged via f55d70f5115df863c776099dc2cad88817adbf84, f5fa697200c236f6278fa18e937e4e2e6cee9537, and bf8bb2c9b1bf49ddb34e77e76707e3c180bf86a5.

daipom:
Thank you for reorganizing commits and merging!

fujimotos:
Background note on why I edited these commits.

One of truly awesome feature of Git is `git bisect`. It allows you to
perform biinary search on Git history to find a bad commit _fast_.

> git-bisect - Use binary search to find the commit that introduced a bug
>
> https://git-scm.com/docs/git-bisect

Now, to let `git bisect` actually usable, we need to keep master history
bisectable -- each commit should be non-breaking.

So while I understand that "test-first" camp people often intentionally
introduce a breaking commit (which I think is fine as a development
methodology), I'd rather keep the public tree bisectable. So I edited em.

PR #393, opened 2022-03-10
Fix a broken test

ashie:
Fluent::NotFoundPluginError is instroduced as of fluentd v1.14.5

Signed-off-by: Takuro Ashie <ashie@clear-code.com>


ashie:
All tests passed.

PR #400, opened 2022-05-25
Add match_regexp configuration option

funtusov:
This PR adds an Input configuration option to selectively download S3 files based on the object key

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -210,8 +210,7 @@ module Fluent::Plugin
               s3 = body["Records"].first["s3"]
               raw_key = s3["object"]["key"]
               key = CGI.unescape(raw_key)
-              match_regexp = Regexp.new(@match_regexp)
-              next unless match_regexp.match?(key) 
+              next unless @match_regexp.match?(key) 
             end
             process(body)
           rescue => e
--- a/test/test_in_s3.rb
+++ b/test/test_in_s3.rb
@@ -673,10 +673,6 @@ EOS
       end
       [message]
     end
-    # d.run(expect_emits: 1)
-    # events = d.events
-    # expected_records = []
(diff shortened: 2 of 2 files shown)


lib/fluent/plugin/in_s3.rb
```diff
@@ -90,6 +90,8 @@ def initialize
     config_param :check_apikey_on_start, :bool, default: true
     desc "URI of proxy environment"
     config_param :proxy_uri, :string, default: nil
+    desc "Optional RegEx to match incoming messages"
+    config_param :match_regexp, :string, default: nil
```
ashie, line 94: Could you use `:regexp` type like this?
https://github.com/fluent/fluentd/blob/f81d60f84cc5fd569c1675f43dcd240cee0a05b6/lib/fluent/plugin/parser.rb#L92

It will check and compile the expression automatically.

lib/fluent/plugin/in_s3.rb
```diff
@@ -204,7 +206,13 @@ def run
             body = Yajl.load(message.body)
             log.debug(body)
             next unless body["Records"] # skip test queue
-
+            if @match_regexp
+              s3 = body["Records"].first["s3"]
+              raw_key = s3["object"]["key"]
+              key = CGI.unescape(raw_key)
+              match_regexp = Regexp.new(@match_regexp)
```
ashie, line 213: `@match_regexp` is already `Regexp` object, you can use it directly so that compilation cost can be reduced.

ashie (changes requested):

lib/fluent/plugin/in_s3.rb
funtusov: Right! 👍 Updated

ashie (approved):

ashie:
Thanks!

PR #401, opened 2022-05-27
Alternative sqs credentials

YannickTeKulve:
Use different aws key id and secret for sqs if supplied

### DIFF
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -282,6 +282,7 @@ module Fluent::Plugin
       if @sqs.aws_key_id && @sqs.aws_sec_key
         options[:access_key_id] = @sqs.aws_key_id
         options[:secret_access_key] = @sqs.aws_sec_key
+      end
       log.on_trace do
         options[:http_wire_trace] = true
         options[:logger] = log
--- a/lib/fluent/plugin/in_s3.rb
+++ b/lib/fluent/plugin/in_s3.rb
@@ -140,6 +140,14 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "sqs/queue_name is required"
       end
 
+      if !!@aws_key_id ^ !!@aws_sec_key
+        raise Fluent::ConfigError, "aws_key_id or aws_sec_key is missing"
+      end
+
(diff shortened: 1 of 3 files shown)


YannickTeKulve:
Hi 

Something I forgot to mention: I chose to overwrite the credentials as not to break backwards compatibility.

YannickTeKulve:
ping

lib/fluent/plugin/in_s3.rb
```diff
@@ -275,6 +279,9 @@ def create_sqs_client
       options[:region] = @s3_region if @s3_region
       options[:endpoint] = @sqs.endpoint if @sqs.endpoint
       options[:http_proxy] = @proxy_uri if @proxy_uri
+      if @sqs.aws_key_id && @sqs.aws_sec_key
+        options[:access_key_id] = @sqs.aws_key_id
+        options[:secret_access_key] = @sqs.aws_sec_key
```
ashie, line 284: `end` is missing

ashie (changes requested):
Sorry, I overlooked it.
Could you check CI results? All of them are failed.

lib/fluent/plugin/in_s3.rb
YannickTeKulve: Sorry cleaned up to much, should work now

lib/fluent/plugin/in_s3.rb
```diff
@@ -275,6 +279,10 @@ def create_sqs_client
       options[:region] = @s3_region if @s3_region
       options[:endpoint] = @sqs.endpoint if @sqs.endpoint
       options[:http_proxy] = @proxy_uri if @proxy_uri
+      if @sqs.aws_key_id && @sqs.aws_sec_key
```
ashie, line 282: Could you raise an error at `configure` when one of them is lacked so that users can noticed their mistake?
https://github.com/fluent/fluent-plugin-s3/blob/fba5e07d1eb7908eb3d6c42abed161e7b8decba2/lib/fluent/plugin/in_s3.rb#L139-L141

ashie (changes requested):

ashie (approved):

ashie:
Thanks!

PR #402, opened 2022-06-14
Support configuring ssl_ca_bundle and ssl_ca_directory in s3 output plugin

chancez:
Used to customize the CA used to verify the S3 API endpoint when working
with S3 compatible APIs using self-signed certificates.

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -97,6 +97,10 @@ module Fluent::Plugin
     config_param :enable_dual_stack, :bool, default: false
     desc "If false, the certificate of endpoint will not be verified"
     config_param :ssl_verify_peer, :bool, :default => true
+    desc "Full path to the SSL certificate authority bundle file that should be used when verifying peer certificates. If unspecified, defaults to the system CA if available."
+    config_param :ssl_ca_bundle, :string, :default => nil
+    desc "Full path of the directory that contains the unbundled SSL certificate authority files for verifying peer certificates. If you do not pass ssl_ca_bundle or ssl_ca_directory the the system default will be used if available."
+    config_param :ssl_ca_directory, :string, :default => nil
     desc "The format of S3 object keys"
     config_param :s3_object_key_format, :string, default: "%{path}%{time_slice}_%{index}.%{file_extension}"
     desc "If true, the bucket name is always left in the request URI and never moved to the host as a sub-domain"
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_s3.rb
```diff
@@ -97,6 +97,8 @@ def initialize
     config_param :enable_dual_stack, :bool, default: false
     desc "If false, the certificate of endpoint will not be verified"
     config_param :ssl_verify_peer, :bool, :default => true
+    desc "Full path to the SSL certificate authority bundle file that should be used when verifying peer certificates. If unspecified, defaults to the system CA if available."
```
ashie, line 100: Could you add this document under [here](https://github.com/fluent/fluent-plugin-s3/blob/e1b206d5d0837bcdd55636f46ffe9bd76b423d3e/docs/output.md#ssl_verify_peer) too?

lib/fluent/plugin/out_s3.rb
```diff
@@ -97,6 +97,8 @@ def initialize
...
+    config_param :ssl_ca_bundle, :string, :default => nil
```
ashie, line 101: It would be better that adding also `ssl_ca_directory`.

ref: https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#initialize-instance_method

ashie (changes requested):

lib/fluent/plugin/out_s3.rb
chancez: Sorry not 100% clear what your asking for, but it sounds like you think we should add both ssl_ca_bundle and ssl_ca_directory? I'm fine with doing both.

lib/fluent/plugin/out_s3.rb
chancez: Updated

lib/fluent/plugin/out_s3.rb
chancez: Okay, I added `ssl_ca_directory` configurations as well.

ashie: Thanks!

ashie (approved):

chancez:
@ashie can we get a 1.8 release with these changes soon potentially?

ashie:
@chancez  I've released v1.7.1.

PR #405, opened 2022-07-21
Fix a typo in IAM Policy section description

ehudon:
## Context

I am currently working with the S3 plugin and noticed a typo in the description of the IAM policy section. I thought I would fix it while I am at it.

## Nature of change

In the example we can see:

```
arn:aws:s3:::my-s3bucket/*
```
This won't match `my-s3bucket-test` but rather `my-s3bucket/test`.


ashie:
Thanks!

PR #423, opened 2023-05-11
GHA: Make staling and closing period shorter

daipom:
It would be good to align these settings with Fluentd.

https://github.com/fluent/fluentd/pull/4116


ashie:
Thanks!

Issue #156, opened 2016-07-19
[warn]: 'aws_iam_retries' parameter is deprecated. Use 'instance_profile_credentials' instead


gozer:
Since aws_iam_retries defaults to '5' in:
  https://github.com/fluent/fluent-plugin-s3/blob/master/lib/fluent/plugin/out_s3.rb#L72

This warning will always be emitted:
  https://github.com/fluent/fluent-plugin-s3/blob/master/lib/fluent/plugin/out_s3.rb#L318

PR #426, opened 2023-07-13
Update documentation in the instance_profile_credentials section

fumihumi:
I asked a question on Discussion (ref: https://github.com/fluent/fluentd/discussions/4236)  the other day, and after examining the commit log I found the following

Although aws_iam_retries is already deprecated, I found the following commit in the commit log for this line.

> config_param :aws_iam_retries, :integer, default: nil, deprecated: "Use 'instance_profile_credentials' instead."

ref: https://github.com/fluent/fluent-plugin-s3/blame/master/lib/fluent/plugin/out_s3.rb#L87

commit: https://github.com/fluent/fluent-plugin-s3/commit/11e8ad76a38d59177bf2ce1435c5bdcdb7f16e67

I thought the above commit was omitting an update to the Document.

### DIFF
--- a/docs/credentials.md
+++ b/docs/credentials.md
@@ -126,7 +126,7 @@ See also:
 
 ### retries
 
-Number of times to retry when retrieving credentials. Default is 5.
+Number of times to retry when retrieving credentials. Default is nil.
 
 ### ip_address
 


ashie:
refs:
* #156 
* #158

You are right, it was changed to suppress deprecation warnings.
Thanks for your contribution!

PR #439, opened 2024-10-18
Add Zstd compression support to S3 plugin

ddukbg:
### Summary

This PR adds support for Zstd compression in the Fluentd S3 plugin.

### Changes

- Implemented Zstd compression using the `zstd-ruby` library.
- Introduced the `ZstdCompressor` class to handle log compression before uploading to S3.
- Updated the example configuration to demonstrate the use of `store_as zstd`.
- Ensured that the `Zstd` module is properly loaded to avoid uninitialized constant errors.

### Testing

- Successfully tested the Zstd compression functionality by sending large log data to S3.
- All test cases passed without errors.

Test Code

```
require 'fluent/test'
require 'fluent/plugin/output'
require_relative '../lib/fluent/plugin/out_s3'  # out_s3.rb 파일을 명시적으로 불러옵니다
require 'fluent/plugin/s3_compressor_zstd'
require 'zstd-ruby'
require 'tempfile'

RSpec.describe Fluent::Plugin::S3Output::ZstdCompressor do
  let(:log) { double('log', warn: nil) }
  let(:compressor) { described_class.new(buffer_type: 'memory', log: log) }

  describe '#compress' do
    let(:test_data) { "This is a test log message" }

    it 'compresses the data using zstd' do
      chunk = double('chunk')
      allow(chunk).to receive(:open).and_yield(StringIO.new(test_data))

      tmp_file = Tempfile.new
      compressor.compress(chunk, tmp_file)
      tmp_file.rewind

      compressed_data = tmp_file.read
      expect(compressed_data).not_to eq(test_data)
      expect(Zstd.decompress(compressed_data)).to eq(test_data)
    end

    it 'logs a warning if compression fails' do
      chunk = double('chunk')
      allow(chunk).to receive(:open).and_raise(StandardError.new("Mock compression error"))

      tmp_file = Tempfile.new
      expect { compressor.compress(chunk, tmp_file) }.to raise_error(StandardError)

      expect(log).to have_received(:warn).with(/zstd compression failed: Mock compression error/)
    end
  end
end
```

Result
```
rspec test/zstd_compressor_spec.rb 

..

Finished in 0.03044 seconds (files took 0.59546 seconds to load)
2 examples, 0 failures
```

store_as (Zstd) Test

```
#fluent.conf
# -*- encoding: utf-8 -*-
<source>
  @type forward
  @id   input
  @label @mainstream
</source>

<label @mainstream>
  <match **>
    @type s3
    s3_bucket fluent-test-yw
    s3_region ap-northeast-2
    path logs/
    store_as zstd
    <format>
      @type json
      time_key timestamp   # JSON의 시간 키 명시
      encoding utf-8       # 인코딩 명시
    </format>
    <buffer>
      @type memory              # 메모리 버퍼 사용
      chunk_limit_size 1m        # 청크 크기를 1MB로 설정 (더 큰 로그를 한번에 처리)
      flush_interval 1s          # 1초마다 플러시
      flush_thread_count 4       # 동시에 플러시할 쓰레드 개수 (시스템 성능에 따라 더 증가 가능)
      retry_max_interval 10s     # 재시도 시간 조정
      retry_timeout 60m          # 재시도 시간 제한
    </buffer>
  </match>
</label>

```

Test Data
`echo '{"message": "'$(head -c 1000000 </dev/zero | tr '\0' 'A')'"}' | fluent-cat test.tag`

fluentd log
`2024-10-18 17:49:37 +0900 [info]: #0 fluent/log.rb:362:info: [Aws::S3::Client 200 0.162773 0 retries] head_object(bucket:"fluent-test-yw",key:"logs/20241018_0.zst")  `

S3 Data
image

### Why this feature?

Zstd compression provides a better compression ratio and performance compared to gzip, making it a valuable option for users who want efficient log storage on S3.


daipom:
Thanks for this enhancement.
Could you please add DCO to all commits?

fluent-plugin-s3.gemspec
```diff
@@ -26,4 +26,5 @@ Gem::Specification.new do |gem|
   # aws-sdk-core requires one of ox, oga, libxml, nokogiri or rexml,
   # and rexml is no longer default gem as of Ruby 3.0.
   gem.add_development_dependency "rexml"
+  gem.add_development_dependency 'zstd-ruby'
```
daipom, line 29: Shouldn't this be `gem.add_dependency`?

Watson1978: Right. I also think it should add gem into runtime dependency instead.

ddukbg: Thank you for your valuable feedback. I have updated the `fluent-plugin-s3.gemspec` file by moving `zstd-ruby` from development dependencies to runtime dependencies as you suggested

daipom:
@ddukbg Thanks! I will review this soon.

FYI: We are just now getting Zstd support on Fluentd's side. 

* https://github.com/fluent/fluentd/pull/4657

ddukbg:
@daipom  hello :)
The unit tests failed on Ruby 2.7, and it seems to be related to an issue with the bundler version. The latest bundler doesn’t appear to support Ruby 2.7. Is there any action I should take to address this?

Watson1978:
ruby/setup-ruby action has been installed proper `bundler` automatically by default.
https://github.com/fluent/fluent-plugin-s3/blob/803cac2094885d5db5d902ab104a883008eda8ec/.github/workflows/linux.yml#L19

image

So, I think you can replace `gem install bundler rake` to `gem install rake` at
https://github.com/fluent/fluent-plugin-s3/blob/803cac2094885d5db5d902ab104a883008eda8ec/.github/workflows/linux.yml#L26

Then, I think the tests might be successful.

lib/fluent/plugin/s3_compressor_zstd.rb
daipom, line 1: Looks like this class is duplicated with `ZstdCompressor` in `out_s3.rb`.
Do we need this file?

test/test_in_s3.rb
daipom, line 1: This test file is for `in_s3`.
We need to add tests to `test_out_s3.rb`.

daipom (changes requested):
Thanks for this enhancement!
Could you please check my following comments?

ddukbg:
@daipom Thank you for your valuable feedback :)
This Pull Request addresses the @daipom feedback regarding the addition of the ZstdCompressor to the S3 output plugin.

**Changes Made:**

1. **Removed Duplicate `ZstdCompressor` Class:**
   - Deleted `lib/fluent/plugin/s3_compressor_zstd.rb` as the `ZstdCompressor` class was already defined in `out_s3.rb`, eliminating duplication.

2. **Added Tests for `ZstdCompressor` in `test_out_s3.rb`:**
   - Moved the tests related to `ZstdCompressor` from `test_in_s3.rb` to `test_out_s3.rb` since they pertain to the output plugin.
   - Added test cases `test_configure_with_mime_type_zstd` and `test_write_with_zstd` to verify the proper functioning of `ZstdCompressor`.

3. **Ensured `ZstdCompressor` is Properly Registered:**
   - Verified that `ZstdCompressor` is correctly registered in the `COMPRESSOR_REGISTRY` within `out_s3.rb`.

4. **Fixed Test Failures:**
   - Adjusted the `s3path` in `test_write_with_zstd` to match the expected file extension for zstd compression, resolving the test failure.

5. **Resolved Gem Build Error:**
   - Removed `s3_compressor_zstd.rb` from the Git index to fix the gem build error caused by the file being listed but not present.


Fluentd Log
```
2024-10-25 18:38:37 +0900 [info]: #0 fluent/log.rb:362:info: [Aws::S3::Client 200 0.119847 0 retries] head_object(bucket:"fluent-test-yw",key:"logs/20241025_0.zst")
2024-10-25 18:38:37 +0900 [info]: #0 fluent/log.rb:362:info: [Aws::S3::Client 200 0.017421 0 retries] head_object(bucket:"fluent-test-yw",key:"logs/20241025_1.zst")
2024-10-25 18:38:37 +0900 [info]: #0 fluent/log.rb:362:info: [Aws::S3::Client 200 0.029599 0 retries] put_object(body:#<Tempfile:/var/folders/59/3d04mk1x5nd_mf9gwd2c1dvw0000gn/T/s3-20241025-86143-etcg6h>,content_type:"application/x-zst",storage_class:"STANDARD",bucket:"fluent-test-yw",key:"logs/20241025_2.zst")
```

S3 Data
Pasted Graphic 8

testcode
```
bundle exec ruby -Ilib:test test/test_out_s3.rb 
Loaded suite test/test_out_s3
Started
Finished in 22.943453 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
53 tests, 76 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------
```

testcode(rspec)
```
rspec test/zstd_compressor_spec.rb                                                                                                                                       

Finished in 0.0169 seconds (files took 0.59358 seconds to load)
2 examples, 0 failures
```

```
require 'fluent/test'
require 'fluent/plugin/output'
require_relative '../lib/fluent/plugin/out_s3'  # out_s3.rb 파일을 명시적으로 불러옵니다
require 'zstd-ruby'
require 'tempfile'

RSpec.describe Fluent::Plugin::S3Output::ZstdCompressor do
  let(:log) { double('log', warn: nil) }
  let(:compressor) { described_class.new(buffer_type: 'memory', log: log) }

  describe '#compress' do
    let(:test_data) { "This is a test log message" }

    it 'compresses the data using zstd' do
      chunk = double('chunk')
      allow(chunk).to receive(:open).and_yield(StringIO.new(test_data))

      tmp_file = Tempfile.new
      compressor.compress(chunk, tmp_file)
      tmp_file.rewind

      compressed_data = tmp_file.read
      expect(compressed_data).not_to eq(test_data)
      expect(Zstd.decompress(compressed_data)).to eq(test_data)
    end

    it 'logs a warning if compression fails' do
      chunk = double('chunk')
      allow(chunk).to receive(:open).and_raise(StandardError.new("Mock compression error"))

      tmp_file = Tempfile.new
      expect { compressor.compress(chunk, tmp_file) }.to raise_error(StandardError)

      expect(log).to have_received(:warn).with(/zstd compression failed: Mock compression error/)
    end
  end
end
```

**Notes:**

- All changes have been made according to the maintainer's comments.
- Tests have been updated and all pass successfully.
- Documentation and code comments have been updated accordingly.

Please review the changes, and let me know if any further adjustments are needed.

lib/fluent/plugin/s3_compressor_zstd.rb
ddukbg: Removed Duplicate ZstdCompressor Class:

Deleted lib/fluent/plugin/s3_compressor_zstd.rb as the ZstdCompressor class was already defined in out_s3.rb, eliminating duplication.

test/test_in_s3.rb
ddukbg: Added Tests for ZstdCompressor in test_out_s3.rb:

Moved the tests related to ZstdCompressor from test_in_s3.rb to test_out_s3.rb since they pertain to the output plugin.
Added test cases test_configure_with_mime_type_zstd and test_write_with_zstd to verify the proper functioning of ZstdCompressor.

lib/fluent/plugin/out_s3.rb
```diff
@@ -630,6 +631,28 @@ def compress(chunk, tmp)
...
+      def content_type
+        'application/x-zst'.freeze
+      end
+
+      def compress(chunk, tmp)
+        uncompressed_data = ''
+        chunk.open do |io|
+          uncompressed_data = io.read
+        end
+        compressed_data = Zstd.compress(uncompressed_data, level: @level)
```
daipom, line 648: > compressed_data = Zstd.compress(uncompressed_data, level: @level)

It appears that `@level` is always `nil`.

daipom, line 648: It could be simpler like this. 

```suggestion
        compressed_data = Zstd.compress(chunk.read, level: @level)
```

test/test_out_s3.rb
```diff
@@ -456,6 +464,33 @@ def test_write_with_custom_s3_object_key_format_containing_hex_random_placeholde
...
+  
+    File.open(s3_local_file_path, 'rb') do |file|
+      compressed_data = file.read
+      uncompressed_data = Zstd.decompress(compressed_data)
+      expected_data = %[2011-01-02T13:14:15Z\ttest\t{"a":1}\n] +
+                      %[2011-01-02T13:14:15Z\ttest\t{"a":2}\n]
+      assert_equal expected_data, uncompressed_data
+    end
+    FileUtils.rm_f(s3_local_file_path)
+  end
```
daipom, line 492: Let's remove unnecessary spaces.

```suggestion
  def test_write_with_zstd
    setup_mocks(true)
    s3_local_file_path = "/tmp/s3-test.zst"

    expected_s3path = "log/events/ts=20110102-13/events_0-#{Socket.gethostname}.zst"

    setup_s3_object_mocks(s3_local_file_path: s3_local_file_path, s3path: expected_s3path)

    config = CONFIG_TIME_SLICE + "\nstore_as zstd\n"
    d = create_time_sliced_driver(config)

    time = event_time("2011-01-02 13:14:15 UTC")
    d.run(default_tag: "test") do
      d.feed(time, { "a" => 1 })
      d.feed(time, { "a" => 2 })
    end

    File.open(s3_local_file_path, 'rb') do |file|
      compressed_data = file.read
      uncompressed_data = Zstd.decompress(compressed_data)
      expected_data = %[2011-01-02T13:14:15Z\ttest\t{"a":1}\n] +
                      %[2011-01-02T13:14:15Z\ttest\t{"a":2}\n]
      assert_equal expected_data, uncompressed_data
    end
    FileUtils.rm_f(s3_local_file_path)
  end
```

daipom (changes requested):
Thanks! Sorry for my late response.
Basically, it looks good to me.
Could you please fix the following?

ddukbg:
@daipom 
Thank you for your helpful suggestions! I've updated the code according to your feedback:
```
- Removed duplicate file reading
- Simplified the compression process
```

Could you please review the changes when you have a chance?

lib/fluent/plugin/out_s3.rb
daipom: @ddukbg What about this point?

ddukbg: @daipom Oh, I forgot to mention.
I'll check that out and comment back.

ddukbg: @daipom 

Hello :)

As per your feedback, I've fixed the issue where `@level` was always `nil`. I've added a `DEFAULT_LEVEL` constant with a value of 3 to the `ZstdCompressor` class and updated the `initialize` method to properly initialize the compression level.

- **Summary of changes:**
  - Added `DEFAULT_LEVEL` constant (value: 3)
  - Fixed compression level initialization
  - Handled line endings and encoding
  - Improved error handling

The commit message is:

```
fix: Add proper compression level handling to ZstdCompressor

- Add DEFAULT_LEVEL constant with value 3
- Fix compression level initialization
- Handle line endings and encoding
- Add proper error handling
```

Please review the updated code.

Thank you.

lib/fluent/plugin/out_s3.rb
daipom, line 1: It would be better to put the implementation of `ZstdCompressor` in a separate file so that `require 'zstd-ruby'` is only executed when necessary.

How about adding such a file instead of fixing `out_s3.rb`?

`s3_compressor_zstd.rb`
```rb
require 'zstd-ruby'

module Fluent::Plugin
  class S3Output
    class ZstdCompressor < Compressor
      S3Output.register_compressor("zstd", self)

      config_section :compress, param_name: :compress_config, init: true, multi: false do
        desc "Compression level for zstd (1-22)"
        config_param :level, :integer, default: 3
      end

      def ext
        'zst'.freeze
      end

      def content_type
        'application/x-zst'.freeze
      end

      def compress(chunk, tmp)
        compressed_data = Zstd.compress(chunk.read, level: @compress_config.level)
        tmp.write(compressed_data)
      rescue => e
        log.warn "zstd compression failed: #{e.message}"
        raise e
      end
    end
  end
end
```

The points are 

* `require 'zstd-ruby'` will be executed only when necessary.
  * This file is automatically loaded when `store_as` is set to `zstd`. (We don't need to change `out_s3.rb`.)
* We can change `level` as follows:
  ```xml
  <match test>
    @type s3
    ...
    <compress>
      level 1
    </compress>
  </match>
  ```
  * Note: We can omit the `compress` section to use the default value `3`.

lib/fluent/plugin/out_s3.rb
```diff
@@ -630,6 +631,36 @@ def compress(chunk, tmp)
...
+        'zst'.freeze
+      end
+    
+      def content_type
+        'application/x-zst'.freeze
+      end
+    
+      def compress(chunk, tmp)
+        begin
+          data = chunk.read.gsub(/\r\n/, "\n").force_encoding('UTF-8')
```
daipom, line 653: > `data = chunk.read.gsub(/\r\n/, "\n").force_encoding('UTF-8')`

Could you please tell me why this is necessary?

test/test_out_s3.rb
```diff
@@ -109,6 +109,14 @@ def test_configure_with_mime_type_lzo
       assert(e.is_a?(Fluent::ConfigError))
     end
 
+    def test_configure_with_mime_type_zstd
+      conf = CONFIG.clone
+      conf << "\nstore_as zstd\n"
+      d = create_driver(conf)
+      assert_equal 'zst', d.instance.instance_variable_get(:@compressor).ext
+      assert_equal 'application/x-zst', d.instance.instance_variable_get(:@compressor).content_type
+    end
```
daipom, line 118: We can add a test for the `level` option like this.

```suggestion
    data('level default' => nil,
         'level 1' => 1)
    def test_configure_with_mime_type_zstd(level)
      conf = CONFIG.clone
      conf << "\nstore_as zstd\n"
      conf << "\n<compress>\nlevel #{level}\n</compress>\n" if level
      d = create_driver(conf)
      assert_equal 'zst', d.instance.instance_variable_get(:@compressor).ext
      assert_equal 'application/x-zst', d.instance.instance_variable_get(:@compressor).content_type
      assert_equal (level || 3), d.instance.instance_variable_get(:@compressor).instance_variable_get(:@compress_config).level
    end
```

daipom (changes requested):
I'm sorry to bother you, but I think it would be better to implement the compressor in a separate file.
Could you please confirm the following comments?

lib/fluent/plugin/out_s3.rb
daipom: https://github.com/fluent/fluent-plugin-s3/pull/439#discussion_r1827143270

> Handled line endings and encoding

Could you please add this test case?

ddukbg: @daipom 
I've found that those lines `(.gsub(/\r\n/, "\n").force_encoding('UTF-8')) `were mistakenly added during my initial troubleshooting of the compression issues. 

The real issue was not with line endings or encoding, but with the compression implementation itself. After fixing the compression logic to use `chunk.write_to` method properly, the compression and tests are now working correctly without any need for explicit line ending or encoding manipulation.

I'll remove those unnecessary lines since they don't serve any purpose in the proper implementation.

ddukbg:
@daipom 
I've found that those lines `(.gsub(/\r\n/, "\n").force_encoding('UTF-8'))` were mistakenly added during my initial troubleshooting of the compression issues. 

The real issue was not with line endings or encoding, but with the compression implementation itself. After fixing the compression logic to use `chunk.write_to` method properly, the compression and tests are now working correctly without any need for explicit line ending or encoding manipulation.

Changes in this commit:
1. Moved ZstdCompressor to a separate file
2. Removed unnecessary line ending and encoding handling
3. Fixed compression implementation using `chunk.write_to`
4. Unit tests are now passing with proper ZSTD compression

I can confirm that the compressed files now have the correct ZSTD magic number (`28 b5 2f fd`) and can be properly decompressed.

lib/fluent/plugin/s3_compressor_zstd.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def compress(chunk, tmp)
+        w = StringIO.new
+        chunk.write_to(w)
+        w.rewind
+        compressed = Zstd.compress(w.read, level: @compress_config.level)
+        tmp.binmode
+        tmp.rewind
+        tmp.write(compressed)
+        tmp.rewind
```
daipom, line 29: Thanks for the fix!
Now, looks good to me except for the `compress` method.

> The real issue was not with line endings or encoding, but with the compression implementation itself. After fixing the compression logic to use `chunk.write_to` method properly, the compression and tests are now working correctly without any need for explicit line ending or encoding manipulation.

On my fork, all `S3OutputTest` tests pass with the following simple implementation.

```suggestion
      def compress(chunk, tmp)
        compressed = Zstd.compress(chunk.read, level: @compress_config.level)
        tmp.write(compressed)
```

I would like to know which cases cannot be handled correctly with this simple implementation.
Do you use Windows?
When I run the test on Windows, the line ending code is different from the expected value.

ddukbg: @daipom 
Thank you for the suggestion! You're right - your simpler implementation works perfectly.

I initially overcomplicated the implementation with unnecessary buffer handling and line ending concerns. Your version is much cleaner and achieves the same result. I tested your simplified version and all tests pass successfully.

I don't use Windows, and my concerns about line endings were unfounded. I was trying to solve compression issues by handling line endings when the real solution was much simpler.

I'll update the code to use your suggested implementation. It's more straightforward and easier to understand.

daipom: Thanks!

daipom (approved):
LGTM! Thanks so much for this enhancement!

fluent-plugin-s3.gemspec
daipom: @ddukbg 
Sorry, I was wrong.
It should be `development_dependency`.
Please take a look at the following PR for details.

* #445

This fix will require manual installation of `zstd-ruby` gem.

PR #447, opened 2024-11-18
out_s3: add debug log for when completed to write chunk

### LINKED ISSUE #414: Add debug log after s3 put
### Is your feature request related to a problem? Please describe.

arch
We'd like to send relatively large size data to Kinesis Data Streams. Due to the price problem, I'm considering send only meta data and body content is sent via S3. To do so, capturing [debug log before putting in S3](https://github.com/fluent/fluent-plugin-s3/blob/2109abcaf6279fb92d0e877359d281aacba6b9ff/lib/fluent/plugin/out_s3.rb#L354), parse it and then send to Kinesis is an idea. But if sending data to S3 takes some time, consumer may fail to fetch body content from S3 because placing in S3 possibly has not been completed and not existed. Can I add debug log after sending S3?

### Describe the solution you'd like

Add log statement

### Describe alternatives you've considered

```
@bucket.object(s3path).put(put_options)

+ log.debug "out_s3: completed put chunk to s3 #{dump_unique_id_hex(chunk.unique_id)} with metadata #{chunk.metadata} to s3://#{@s3_bucket}/#{s3path}"
```

### Additional context

_No response_

Watson1978:
Fix #414

This patch will add debug log for when completed to write chunk to S3.

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -378,6 +378,8 @@ module Fluent::Plugin
         end
         @bucket.object(s3path).put(put_options)
 
+        log.debug "out_s3: completed to write chunk #{dump_unique_id_hex(chunk.unique_id)} with metadata #{chunk.metadata} to s3://#{@s3_bucket}/#{s3path}"
+
         @values_for_s3_object_chunk.delete(chunk.unique_id)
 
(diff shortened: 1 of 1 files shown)


daipom (approved):
LGTM. Thanks!

Issue #416, opened 2023-01-27
S3 output filling up /tmp


waqarsky:
### Describe the bug

I have these files on /tmp filling up disk space and they are not removed by fluentd

```
30M	s3-20230115-3429444-1n0gm68
34M	s3-20230118-4051437-8mi5ts
36M	s3-20230105-1633445-ymzdql
44M	s3-20230118-4082680-1v29q88
56M	s3-20230111-2775877-5b3vmz
63M	s3-20230109-2467022-n12d0g
65M	s3-20230121-437995-1sq1mt9
75M	s3-20230106-1845823-cd9qij
76M	s3-20230111-2738261-1a3zss4
77M	s3-20230108-2205141-scyh2z
83M	s3-20230116-3580308-le9f4y
98M	s3-20230108-2273951-1fwklwx
132M	s3-20230119-4177644-1b82v95
151M	s3-20230109-2341735-1p706j2
198M	s3-20230120-229601-t0fspm
232M	s3-20230113-3179791-qhwem0
244M	s3-20230105-1555034-fj7pmg
244M	s3-20230110-2699071-y6dgc7
244M	s3-20230112-3003537-m4vg83
244M	s3-20230117-3937170-1s4izqw
```

### To Reproduce

Set up the plugin with the config below. These files are very infrequent, not all of the ones created remain. So the plugin leaves some in /tmp and deletes most as it should  

### Expected behavior

tmp files should be removed after upload or there should be an error stating why the files are still there

### Your Environment

```markdown
- Fluentd version: fluentd (1.15.2)
- fluent-plugin-s3 version: 1.7.1
- Operating system
- Kernel version: 
5.15.0-1022-aws #26~20.04.1-Ubuntu
```

### Your Configuration
```
  </store>
  <store ignore_error>
    @type s3
    @log_level info
    s3_bucket "BUCKET"
    s3_region "REGION"
    acl bucket-owner-full-control
    store_as "json"
    <format>
      @type json
    </format>
    use_server_side_encryption "ENCRYPTION"
    path %Y/%m/%d/
    time_slice_format %Y%m%d-%H%M%S
    s3_object_key_format %{path}%{uuid_flush}-%{hex_random}-%{time_slice}-%{hms_slice}.%{file_extension}
    check_object false
    slow_flush_log_threshold 100.0
    <buffer REDACTED>
      retry_wait 20s
      retry_exponential_backoff_base 2
      retry_type exponential_backoff
[...]

PR #448, opened 2024-11-20
out_s3: add log to output unexpected error of Tempfile#close

Watson1978:
It is difficult to take any action deal with as there was no output if Tempfile#close caused unexpected error. So, this patch will add a log to know that an error has occurred in Tempfile#close.

Related to #416


daipom (approved):
LGTM. Thanks!!

Issue #451, opened 2025-01-23
Allow set http proxy and sts_endpoint_url for web_identity_credentials


rubroboletus:
### Is your feature request related to a problem? Please describe.

We are using specific VPC Endpoint for STS and web identity credentials in our Kubernetes cluster on AWS. Currently we are unable to use this STS endpoint and must set a global http proxy for all the fluentd pod.

### Describe the solution you'd like

We like to use our VPC Endpoint for STS with web identity credentials.

### Describe alternatives you've considered

Currently we are using a global http proxy settings and no_proxy settings for all the fluentd pod.

PR #452, opened 2025-01-23
add sts_http_proxy and sts_endpoint_url to web_identity_credentials

rubroboletus:
possibly solves https://github.com/fluent/fluent-plugin-s3/issues/451

### DIFF
--- a/lib/fluent/plugin/out_s3.rb
+++ b/lib/fluent/plugin/out_s3.rb
@@ -62,6 +62,10 @@ module Fluent::Plugin
       config_param :duration_seconds, :integer, default: nil
       desc "The region of the STS endpoint to use."
       config_param :sts_region, :string, default: nil
+      desc "A http proxy url for requests to aws sts service"
+      config_param :sts_http_proxy, :string, default: nil, secret: true
+      desc "A url for a regional sts api endpoint, the default is global"
+      config_param :sts_endpoint_url, :string, default: nil
     end
     config_section :instance_profile_credentials, multi: false do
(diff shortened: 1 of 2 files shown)


daipom:
Could you add DCO?

daipom:
@rubroboletus Thanks for this enhancement!!
So, this PR adds the same options to `web_identity_credentials` as `assume_role_credentials`.
It looks good to me.

If you can, could you please add a note to the next document about these options?

* https://github.com/fluent/fluent-plugin-s3/blob/master/docs/credentials.md

Since there is no description of these existing options in `assume_role_credentials`, it would be helpful if you could add a description for both `assume_role_credentials` and `web_identity_credentials`.

rubroboletus:
> Could you add DCO?
> 
> https://github.com/fluent/fluent-plugin-s3/pull/452/checks?check_run_id=36048644495

added

rubroboletus:
> @rubroboletus Thanks for this enhancement!! So, this PR adds the same options to `web_identity_credentials` as `assume_role_credentials`. It looks good to me.
> 
> If you can, could you please add a note to the next document about these options?
> 
> * https://github.com/fluent/fluent-plugin-s3/blob/master/docs/credentials.md
> 
> Since there is no description of these existing options in `assume_role_credentials`, it would be helpful if you could add a description for both `assume_role_credentials` and `web_identity_credentials`.

Sorry, but I'm not a right one person for this task. My english is on very basic level.

daipom:
Thanks for this enhancement!

PR #478, opened 2026-06-10
Replace yajl-ruby with the json gem for JSON handling

Watson1978:
Since the yajl-ruby gem depends on a deprecated Ruby C API, it cannot be installed with Ruby 4.1.


Watson1978:
No remaining Yajl in codes.
```
$ grep -i "yajl" -r *
ChangeLog:  * Use Yajl instead of to_json not to raise exceptions when it got invalid bytes as UTF-8.
```

kenhys (approved):
LGTM.

